PR #16, opened 2024-03-06
fix: make ldap connections persistent and restartable

j-hoff:
previously the request_local_store was used, so new ldap connections were opened per request.
this was changed to a single persistent connection by using app_local_store. which then needs to be restartable.
there is a bug in python-ldap 2.8.1 which causes restartable connections to fail, so we bump to 2.9.1


fritterhoff:
Ah. That could explain the pretty high ldap load in our productive env....

fritterhoff (approved):

PR #18, opened 2024-03-06
feat: smstoken: add configurable URL which is called after check, extend test

j-hoff:
Introduce new parameter POST_CHECK_URL which, when set in the SMS gateway definition, is called after validate/check on the token.
This can be used to configure a web hook which clears a TAN message from a device after it was entered.


fritterhoff (approved):
LGTM

PR #28, opened 2024-03-15
docs: cleanup old refs and fix path names

j-hoff:
removed some more refs to old versions and changed eduMFA to lowercase in file or path names


fritterhoff (approved):
LGTM!

PR #29, opened 2024-03-15
fix: rename provider module names in DB on migration

j-hoff:
in table *smsgateway* the column *providermodule* may contain module names like `privacyidea.lib.smsprovider.HttpSMSProvider.HttpSMSProvider`, rename those to `edumfa.lib.xxx`


fritterhoff (approved):
LGTM 👍

PR #32, opened 2024-03-19
feat: token janitor find by user

pmainz:
This pull request adds the possibility to filter tokens by user in edumfa-token-janitor.
The username should be given in the format 'user@realm'. In case no realm is given the default realm is used.


fritterhoff (approved):
👍🏼

PR #36, opened 2024-03-20
feat: api: set tokeninfo on init

j-hoff:
in api function token/init/ add optional parameter **"info"**, which can contain a json object whose key-value pairs will be set as tokeninfo for the new token.
this might significantly reduce the amount of requests needed in some scenarios.


pmainz (approved):
looks good! :)

PR #54, opened 2024-04-12
fix: correct installation of psycopg2-binary and galera bug

fritterhoff:


j-hoff (approved):
nice work

j-hoff (approved):
yes, probably better

PR #87, opened 2024-04-16
Update documentation (no corresponding issue)

Thaoden:
Some links were broken, they have been cleaned up.

Also, the naming scheme of the primary branch was not used consistently (`main` vs `master`), this too has been cleaned up.


j-hoff (approved):
thank you!

PR #89, opened 2024-04-19
fix: print logo to stderr

Johnnynator:
stdout should be preserved for well formated output that can be properly
processed by other tools.
This simplifies using the output of edumfa-manage in e.g. Ansible.

This was also the behaviour of privacyIDEA.


fritterhoff (approved):

fritterhoff:
Thanks for the correction. Will release it with the next (bugfix) release next week.

PR #108, opened 2024-04-24
feat: add support for PostgreSQL in backup script

Johnnynator:
- **fix: use correct namespace of make_url**
- **feat: add support for PostgreSQL in backup script**
- **fix: remove unused import in backup.py**


fritterhoff (approved):
Sadly I don't have postgres version of eduMFA running right now. Using pen and paper that looks fine for me.

PR #217, opened 2024-07-18
Lint code with black, isort, flynt and pyupgrade

Luc1412:
This pull requests formats the code with:
- black (A PEP 8 compliant code formatter enforcing the black code style)
- isort (Import sorting)
- flynt (standardize formatted strings)
- pyupgrade (upgrade syntax to current min py version)

Additionally a CI check for pull requests has been added, that checks if the code is compliant with black.


edumfa/api/lib/utils.py
```diff
@@ -287,17 +299,23 @@ def get_all_params(request):
...
         # Add the unquoted HTML and form parameters
         return_param = check_unquote(request, request.values)
 
     if request.method == "GET" and request.form:
         return_param.update(check_unquote(request, request.form))
 
     if request.is_json:
-        log.debug("Update params in request {0!s} {1!s} with JSON data.".format(request.method,
-                                                                                request.base_url))
+        log.debug(
```
fritterhoff, line 311: Seems like flynt did not catch everything...

edumfa/lib/token.py
```diff
@@ -2465,32 +2670,40 @@ def get_dynamic_policy_definitions(scope=None):
...
-        conf = get_tokenclass_info(ttype, section='user')
-        if 'enroll' in conf:
-            pol[SCOPE.USER]["enroll{0!s}".format(ttype.upper())] = {
-                'type': 'bool',
-                'desc': _("The user is allowed to enroll a {0!s} token.").format(ttype.upper()),
-                'mainmenu': [MAIN_MENU.TOKENS],
-                'group': GROUP.ENROLLMENT}
+        pol[SCOPE.ADMIN][f"enroll{ttype.upper()!s}"] = {
+            "type": "bool",
+            "desc": _("Admin is allowed to initialize {0!s} tokens.").format(
```
fritterhoff, line 2677: Can we fix some expression like this one? could imagine that we have to encapsulate the f-string in `()` and then apply the `_` translation?
Think we also have to check the translations...

fritterhoff (changes requested):
Looks good so far... I think we should especially recheck the formatings and the prop. the translations must be generated again.

edumfa/lib/token.py
Luc1412: When it comes to localization, we need to keep using `.format`, otherwise it won't work.

fritterhoff: Ok. Thanks.

edumfa/api/lib/utils.py
Luc1412: Yeah unfortunately it didn't catch everything. About 400-500 possible occurrences of old string formatting are still present. Not sure if it's worth to update it manually.

fritterhoff (changes requested):
@Luc1412 I think it would make sense to exclude the most of the commits from git blame like it is explained here: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view ?

Could you also (re)generate the translation files?. The most line numbers will have changed.

Luc1412:
> @Luc1412 I think it would make sense to exclude the most of the commits from git blame like it is explained here: https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view ?

Isn't that solved by just using Squash Merge or do I miss the point here?

> Could you also (re)generate the translation files?. The most line numbers will have changed.

Is there any documentation on how to do that?

PR #218, opened 2024-07-18
chore: migrate static metadata to `pyproject.toml`

Luc1412:
This PR streamlines the package metadata including dependencies. Following changes have been made:
- Static metadata has been migrated from the `setup.py` to a new `pyproject.toml`.
- extra dependencies (for test, docs, posgres, hsm and kerberos) have been updated in the `pyproject.toml`. Associated `requirements.txt` files for these extra requirements have been removed.


fritterhoff:
Great! Thanks for the attempt. I am not sure if we will get rid of the old setup.cfg since the used `data_files` has no version in pyproject.toml ...

Luc1412:
> Great! Thanks for the attempt. I am not sure if we will get rid of the old setup.cfg since the used `data_files` has no version in pyproject.toml ...

The focus is more on static metadata. Having a setup.py is neither deprecated nor an issue.

pyproject.toml
```diff
@@ -0,0 +1,89 @@
...
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12"
+]
+dynamic = ["dependencies"]
+
+[project.urls]
+Homepage = "https://edumfa.io/"
+Documentation = "https://edumfa.readthedocs.io/en/v2.1.0/"
```
fritterhoff: Maybe we can change that to a more generic link so we don't have to update it everytime?

fritterhoff (changes requested):
Looks good so far. Beside the version in the url I think we can merge it next week? :)

fritterhoff (approved):

PR #288, opened 2024-08-23
fix: TOTP and HOTP help strings

Johnnynator:
The omission of privacyIDEA Authenticator as the first app resulted in
naming the "other" Apps as "other" kinda nonsensical.

Adapted to be closer to the current German "Translation"

The question is, do we want to keep a List of other Apps in there? The eduMFA Authenticator App doesn't seem to be Production ready yet.

Most translations right now refer to Google Authenticator


fritterhoff:
Hi,

well. The eduMFA Authenticator will not support classic TOTP tokens. There are several other - better - apps that support such fancy stuff like autocompletion, (cloud) backups, revealing the secrets, custom icons and so on. Thats why we decided that we won't support this sort of token in our app and focus on Push Tokens.

Flo

Johnnynator:
Okay, so I will also change the German translation on Monday. Should there be some Apps listed in the help atring by default?

fritterhoff:
Since there are hunderts of apps I think it would make sense to keep it as neutral as possible.

Johnnynator:
I would go forth with the following two texts (and the equivalent for TOTP).
Any remarks before I go forth and fight msgmerge to not mix up HOTP and TOTP strings again.

English:
> The HOTP token is an event based token. You can paste a secret key or
    have the server generate the secret and scan the QR code with an Authenticator
    App turning your smartphone into an authentication device.   
    But note, that some Apps might have limitations in the supported hash algorithms or other parameters.

German:
> Der HOTP-Token ist ein ereignisbasierter Token. Sie können den geheimen OTP-Schlüssel hier einfügen
   oder den Server einen Schlüssel generieren lassen.
   Diesen können Sie in Ihre Authenticator-App importieren, indem Sie den QR-Code scannen.
   Beachten Sie, dass nicht alle Authenticator-Apps möglicherweise alle Parameter unterstützen.

fritterhoff:
Looks good to me.

Johnnynator:
Pushed changes, also updates some other minor translations. It seems like the `edumfa/translations`  did not get regenerated lately and were missing some changes.

fritterhoff (approved):

PR #289, opened 2024-08-23
feat: change import filename arguments to type click.File

Johnnynator:
All command line flags with the same name should have the same behavior for consistency’s sake.
Using `click.File` gives us the option to read from `stdin` for free when the user specifies `-`.

The same could be done for other flags.


fritterhoff (approved):
LGTM

PR #294, opened 2024-08-27
docs: document eduPUSH type and deprecate Push Token Type

### LINKED ISSUE #278: Document edupush and (legacy) push token types

Currently only Legacy Push (as Push) token type is documented.

Since there is a new Edupush token type, documentation is required. Furthermore the current push token type needs to be renamed and marked as legacy.

Luc1412:
This pull request updates documentation regarding:
- EduPush Token Type
- Mark old push token type as legacy and provides deprecation note

### DIFF
--- a/doc/configuration/sms_gateway_config.rst
+++ b/doc/configuration/sms_gateway_config.rst
@@ -29,11 +29,11 @@ via the Google Firebase service and this is used for :ref:`edupush_token` and :r
 
 
 You can get the necessary *JSON config file*, from your Firebase console.
-The Push authenticator App (eduMFA Authenticator) which you can find in Google Play Store and
+The eduPUSH authenticator App (eduMFA Authenticator) which you can find in Google Play Store and
 Apple App Store uses a Firebase project, that is managed by the GWDG.
 You need to get in touch trough support@gwdg.de with them, to receive a JSON config file for accessing the project.
 
-When using the the legacy Push token type, please refer to the privacyIDEA documentation. However we don't recommend or support using it anymore.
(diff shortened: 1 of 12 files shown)


Luc1412:
@fritterhoff How do we want to communicate the change? 
- Do we want to call our new edupush token type `Push Token` or do we want to refer `EduPush Token` or `eduPush Token` or `Edupush Token`.
- Depending on the first decision, how do we want to refer to the old Push Token type? `Legacy Push Token` or simply `Push Token`
- If there are mentions of Push Token types, how do we want to refer to them? Should we mention both, just Edupush or should we summarize them as Push Token types?

Here also some special cases that needs to be clarified:
### SMS Gateway Config
There it mentions the setup of firebase. The old docs refer to SLA with NetKnights to use the privacyIDEA app. 

Now it's managed by GWDG, but how do people get access to our Firebase instance?
### WebUI Config
There are `show_android_privacyidea_authenticator` and `show_ios_privacyidea_authenticator` we didn't replaced yet. What happens to these options?
### Supported Tokens - Apps
There it mentions privacyIDEA Authenticator for HOTP, TOTP and PUSH. Should we keep it or remove it?

fritterhoff:
* I would suggest `eduMFA Push Token`
* We should use `Legacy Push Token`
* Right now I would summarize them

Right now I check with the GWDG whats the preferred way to get access.

I think we should replace/extend the existing options... Maybe in the same PR or in a new. 

Regarding HTOP, TOTP -> https://github.com/eduMFA/eduMFA/pull/288

fritterhoff:
The preferred way is to write an email to support@gwdg.de

fritterhoff (approved):

PR #322, opened 2024-09-18
POC: webauthn registration split thingie

### LINKED ISSUE #320: Split WebAuthn and Passkey registration in the UI
Right now one needs to set `webauthn_resident_key` to `required` to get some systems to reliably write them on e.g. YubiKeys.
At least on Windows Setting it to `preferred` Results in no Resident Key written, only if set to `required`.
With Firefox on Linux it kinda works.

Splitting both in the Registration UI to give the User the Option to Register either a Passkey or a "Non-resident" Token would be nice.

I only don't see a straight forward way to implement this (without adding a new Token Type).

Johnnynator:
Just some quick POC I hacked together as a basis for discussion.

There seem to be two ways to get this going in an easy way
a) Expand the API to allow sending some kind of webauthn type on the first init call (done in this poc)
b) Do it purely in Frontend and Patch the returned WebAuthn Credential Options there

TODO:
- [ ] Policy to enable/disable the user facing token selection
- [ ] UI Dependent on `AUTHENTICATOR_RESIDENT_KEY` set to `preferred`
- [ ] Also Require User Presence for Passkey type? (Or is this already implicit by requiring a resident key?)
- [ ] Flesh out UI, some Description Text

Resolves #320

### DIFF
--- a/edumfa/api/token.py
+++ b/edumfa/api/token.py
@@ -163,6 +163,7 @@ def init():
     :jsonparam rollover: Set this to 1 or true to indicate, that you want to rollover a token.
                     This is mandatory to rollover tokens, that are in the clientwait state.
     :jsonparam info: string representation of a json object whose key-value pairs will be set as tokeninfo.
+    :jsonparam webauthn_tyoe: String representing a webauthn type (resident, non-resident)
 
     :return: a json result with a boolean "result": true
 
--- a/edumfa/lib/tokens/webauthntoken.py
+++ b/edumfa/lib/tokens/webauthntoken.py
@@ -94,6 +94,7 @@ it in two steps:
 
     type=webauthn
     user=<username>
+    webauthntype=<webauthnType>
 
 The request returns:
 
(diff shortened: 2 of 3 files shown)


fritterhoff:
Mhm. From my point of view it would make the most sense to handle it in the UI only. This is quite cumberstone overall and from my point of view changes pretty often. We started to test different browsers and OS and noticed plenty of different effects...

Johnnynator:
> We started to test different browsers and OS and noticed plenty of different effects...

I see mainly two things that are broken right now, either
a) not having a Passkey written to e.g. a YubiKey
or
b) not being able to Older Tokens that don't support Passkeys/Client-side discoverable Credentials

The Registration API for Tokens only allows forcing Resident Keys, not the other way around.
So for ResidentKeyRequirement `discouraged` and `preferred` is per Documentation somewhat the same (with only different recommend priority) and only `required` is strictly different.

Setting `ResidentKeyRequirement` to `discouraged`/`preferred` still allows writing a Passkey optionally (which Systems that only Support Passkeys like Passwort Managers or other Software Tokens always do).
We can only kinda control what Hardware Tokens do since most other things seem to implement one single way of doing things.

If there are other things that should be considered, handled, then please mention them. My main goal right now is being able to use "legacy" Hardware Tokens without breaking using newer Tokens for Passkeys.

fritterhoff:
Hi and sorry for the little delay.

Regarding this problem. In my opinion, we are trying to fix some unpleasant behavior from the manufacturers/vendors. I understand this problem and we also were facing this problem for our rollout and solved it with some client-side fixes/adaptions. 

I would like to try to keep this intervention to a minimum but I support any solution that improves this flow 😁 
In any case, we should maintain (backward) compatibility with the old APIs and not introduce any breaking changes.

PR #338, opened 2024-09-29
feat: add option for setting fcm push timeout

### LINKED ISSUE #275: Timeout option for push token
The FCM requests are blocking the complete `/validate/check` flow right now. Since that could cause some strange issues in case of missing internet connection we should add a timeout option for this requests.

fritterhoff:


fritterhoff:
Problem gets solved partially. Timeout works but still causes a status 500. Would be better to simply generate no challenge at all.

fritterhoff:
Must additionally reduce the retry amount.

PR #405, opened 2024-11-01
Add content-available and client options

oculos:
This allows processing a notification request without asking the server if there are pending notifications.

With `content-available : 1`, the application wakes up and processes the request without user intervention.

This also allows the user to approve/deny a notification without opening the app (actionable notifications), including on the Apple Watch, if the application supports it.

I also added the possibility to send one key/value called "client", so that calling applications can send a custom text to eventually be displayed by the App.

Changes: 

- Added "content-available : 1" to allow immediate processing of the notification, without the need to contact the server
- Added "time-sensitive" so that notifications get priority, due their sensitive nature
- Added checking for client parameter to include on the push notification, so that apps can use it.


oculos:
Hi! I wonder what's the status of this PR?

Luc1412 (commented):
I think it's finally time to address this PR. First, could you please update the branch and update the code so all checks pass. 

Then regarding your changes:
- `content-available`: I think this change isn't needed. Here the docs regarding that:
> Background update notifications improve the user experience by giving you a way to wake up your app periodically so that it can refresh its data in the background. When apps do not run for extended periods of time, their data can become outdated. When the user finally launches the app again, the outdated data must be replaced, which can cause a delay in using the app. A background update notification can alert the user or it can occur silently.

We don't need a background updates. We only want to show a simple notification which doesn't require this option.
- `interruption-level`: Good change
- Addition of `client`, `user` and `realm`: While it's not yet implemented in the app, it sounds like a reasonable change. Did you test what data exactly gets passed? Could you provide an example for all of these.

oculos:
> I think it's finally time to address this PR. First, could you please update the branch and update the code so all checks pass.

Thanks, will do so, but let's first reach an agreement about the things below: 

> Then regarding your changes:
> 
> * `content-available`: I think this change isn't needed. Here the docs regarding that:
> 
> > Background update notifications improve the user experience by giving you a way to wake up your app periodically so that it can refresh its data in the background. When apps do not run for extended periods of time, their data can become outdated. When the user finally launches the app again, the outdated data must be replaced, which can cause a delay in using the app. A background update notification can alert the user or it can occur silently.
> 
> We don't need a background updates. We only want to show a simple notification which doesn't require this option.

If I remember correctly (there's almost a year since I worked with this), the idea was:

- to already prepare the application to display the interface. On the official EduMFA app, you need to refresh to see if there's a challenge, but with Content-Available, the app does that in background, so when the user opens the app it is already displaying the right interface.
- I think - and right now I am not so sure - it was what made it possible to use native interactions without opening the app:

Screenshot 2025-10-14 at 12 48 22

As you see in the picture above, when Content-Available is sent, the application, when showing the notification, can be programmed to have actionable buttons which can be used to send answers to the push notification without opening the app - Content-Available awakes the app and processes this. I can send you the code I wrote to handle this if you want. The advantage was that one can react to the notifications on both the device, as well as on the Apple Watch or on a Mac, as you see above.

I would suggest you to try to enable it and test - if it doesn't break it, I think it offers a lot of possibilities.

> * `interruption-level`: Good change
> * Addition of `client`, `user` and `realm`: While it's not yet implemented in the app, it sounds like a reasonable change. Did you test what data exactly gets passed? Could you provide an example for all of these.

Yes. The client information is something you send by the application who requests the authentication. The app then shows it to show the user where the login is coming from: 

Screenshot 2025-10-14 at 12 52 36

II don't remember why I felt the need to send realm and user. I checked now on my code and there was no use of those attributes on the notification. I wonder if I thought I'd use them for something else but I really don't remember.

Luc1412:
> to already prepare the application to display the interface. On the official EduMFA app, you need to refresh to see if there's a challenge, but with Content-Available, the app does that in background, so when the user opens the app it is already displaying the right interface.

I think it isn't related to that. Here what I did:
1. I got eduMFA latest and eduMFA Authenticator latest installed on IOS 26
2. I force closed the eduMFA Authenticator App
3. I triggered a challenge
4. My iPhone got the notification, which I clicked
5. The app was opened and the prompt was shown immediately without manually refreshing

There are cases where the app didn't showed a prompt. I couldn't find any consistent way to reproduce it. Did you test `content-available` with the latest version? If so, we could consider adding that.

> As you see in the picture above, when Content-Available is sent, the application, when showing the notification, can be programmed to have actionable buttons which can be used to send answers to the push notification without opening the app - Content-Available awakes the app and processes this. I can send you the code I wrote to handle this if you want. The advantage was that one can react to the notifications on both the device, as well as on the Apple Watch or on a Mac, as you see above.

This feature shouldn't be related to `content-available`. Notification actions don't need to be processed, instead they need to be registered within the app. 

> I would suggest you to try to enable it and test - if it doesn't break it, I think it offers a lot of possibilities.

It doesn't breaks anything, but also seem to not any advantage to our use-case, based on my research.

> Yes. The client information is something you send by the application who requests the authentication. The app then shows it to show the user where the login is coming from:

Do you got some data scheme or a JSON on whats included?

oculos:
> > to already prepare the application to display the interface. On the official EduMFA app, you need to refresh to see if there's a challenge, but with Content-Available, the app does that in background, so when the user opens the app it is already displaying the right interface.
> 
> I think it isn't related to that. Here what I did:
> 
> 1. I got eduMFA latest and eduMFA Authenticator latest installed on IOS 26
> 2. I force closed the eduMFA Authenticator App
> 3. I triggered a challenge
> 4. My iPhone got the notification, which I clicked
> 5. The app was opened and the prompt was shown immediately without manually refreshing
> 
> There are cases where the app didn't showed a prompt. I couldn't find any consistent way to reproduce it. Did you test `content-available` with the latest version? If so, we could consider adding that.
> 
> > As you see in the picture above, when Content-Available is sent, the application, when showing the notification, can be programmed to have actionable buttons which can be used to send answers to the push notification without opening the app - Content-Available awakes the app and processes this. I can send you the code I wrote to handle this if you want. The advantage was that one can react to the notifications on both the device, as well as on the Apple Watch or on a Mac, as you see above.
> 
> This feature shouldn't be related to `content-available`. Notification actions don't need to be processed, instead they need to be registered within the app.
>
> > I would suggest you to try to enable it and test - if it doesn't break it, I think it offers a lot of possibilities.
> 
> It doesn't breaks anything, but also seem to not any advantage to our use-case, based on my research.

No, you see, the test you did wouldn't work anyway, because you need a proper delegate on your code to handle the application awakening. That delegate is `[application(_:didReceiveRemoteNotification:fetchCompletionHandler:)](https://developer.apple.com/documentation/UIKit/UIApplicationDelegate/application(_:didReceiveRemoteNotification:fetchCompletionHandler:))`.

The whole point was to avoid polling the server. When you have Content-Available, your application don't need to poll the server: it already knows you need to display the interface. If I remember well, the official app polls the server every time it becomes active. I considered implementing polling as well - and we might do that - but it seemed a waste to poll when the app already knew - via Content-Available - there there is a challenge to reply. 

So when the user opens the app, the app already has the right interface to show. 

I don't know how Flutter works, but on native iOS, if the you receive a message while the application isn't in foreground, the app doesn't do anything when you open it. It doesn't even know it received a notification. Then you have to poll the server. But with Content-Available, your app knows already that a notification was sent, you can do something about right away.

> > Yes. The client information is something you send by the application who requests the authentication. The app then shows it to show the user where the login is coming from:
> 
> Do you got some data scheme or a JSON on whats included?

Yes. As you see in the server code, there's usually a tendency forward all parameters to the application, but in some methods it doesn't do that. 

On the original code for the radius plugin for PrivacyIDEA, he sent "client=xxx" as a query parameter:

https://github.com/privacyidea/FreeRADIUS/blob/2ec2b17134550cf8f8612d2c73bec72aedad42a8/privacyidea_radius.pm#L494 (note that he also sent user and client)

In my code to trigger a challenge, I'd send: 

```
JSONObject response = sendRequest("get","/validate/triggerchallenge?user="+user+"&client="+client,"", true);
```

EduMFA takes whatever parameters you send as a query parameter. So, for example:

```
https://myedufa.com/validate/triggerchallenge/?user=myuser&client=myapplication
``` 

Is normally accepted by EduMFA.

However, EduMFA was not sending all the other attributes to the app when push token was used, even though it had the "extra_data" parameter on the `create_push_url` function. 

So I'm not sure why the author of PrivacyIDEA sent "client=xxx" on the radius code, but didn't have code to decode that.

oculos:
Hi Lukas,

Let me know if you want me to remove the «content-Available» (I’ll patch our instance to use it). 
then I can rebase my code to make it up to date.

Luc1412:
Sorry for late response. I think we can add it for now, but I'll re-evalute it with the current App migration if it's necessary or not.

oculos:
> Sorry for late response. I think we can add it for now, but I'll re-evalute it with the current App migration if it's necessary or not.

Thanks @Luc1412 ! Can this be merged soon?

Luc1412:
@oculos I evaluate the changes I need to make for the new app I'm currently working on. All I've done is implemented in #981. When this is merged we can discuss which remaining changes we can implement for the app. But I'll plan on finishing it with the next release.

Issue #360, opened 2024-10-08
docker compose fails with database errors: sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) index ix_token_serial already exists


kastl-ars:
### Top-level intent
Running eduMFA via docker compose, as stated in the docs, fails.

### Steps to reproduce

0. Create a new directory and change to it
1. Create a `config.cfg` file according to [this documentation](https://edumfa.readthedocs.io/en/v2.3.0/installation/system/inifile.html)
2. Create a docker-compose.yml fle following the [documentation](https://edumfa.readthedocs.io/en/v2.3.0/installation/docker.html).
3. Run `docker compose up`

### Expected outcome

I would have expected the documentation example to just work.

### Actual outcome

This is the full log. The errors regarding `EDUMFA_LOGCONFIG` can be solved by setting it properly in the configuration file (see #359).

```
docker compose up                                                 
[+] Running 3/3                                        
 ✔ Network docker-compose_default       Created                                                                                                                                                                                                         0.0s 
 ✔ Volume "docker-compose_edumfa_data"  Created                                                                                                                                                                                                         0.0s 
 ✔ Container docker-compose-edumfa-1    Created                                                                                                                                                                                                         0.1s 
Attaching to edumfa-1                                                                                                         
edumfa-1  | Can not import grpc modules.                                                                                      
edumfa-1  |                                                                                                                   
edumfa-1  |                _       __  __ ______                                                                              
edumfa-1  |               | |     |  \/  |  ____/\                                                                            
edumfa-1  |        ___  __| |_   _| \  / | |__ /  \                                                                           
edumfa-1  |       / _ \/ _` | | | | |\/| |  __/ /\ \                                                                          
edumfa-1  |      |  __/ (_| | |_| | |  | | | / ____ \                                                                         
edumfa-1  |       \___|\__,_|\__,_|_|  |_|_|/_/    \_\       v2.3.0
edumfa-1  |                                                                                                                   
edumfa-1  | Could not use EDUMFA_LOGCONFIG: expected '<document start>', but found '<scalar>'                                 
edumfa-1  |   in "<unicode string>", line 4, column 1:     
edumfa-1  |     keys=detail                                                                                                   
edumfa-1  |     ^                                
edumfa-1  | Encryption key written to /etc/edumfa/enckey
edumfa-1  | The file permission of /etc/edumfa/enckey was set to 400!                                       
[...]

PR #448, opened 2024-11-28
fix: stamp db before migration in docker

fbmei:
### Description

I have got the same error with the docker container as issue #360. As far as i understand the docker-setup.sh script tries to upgrade the new database before it has been stamped. I don't know if this is meant to be like this and if there is a deeper error in the code but i changed the docker-setup.sh to check for the stamp of the database before upgrading and its working.

### DIFF
--- a/deploy/docker-setup.sh
+++ b/deploy/docker-setup.sh
@@ -17,6 +17,14 @@ fi
 # Create DB
 echo "Creating DB"
 edumfa-manage createdb
+
+# Check and stamp DB
+STAMP=$(edumfa-manage db current -d /usr/local/lib/edumfa/migrations 2>/dev/null)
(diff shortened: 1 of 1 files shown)


fritterhoff:
I am wondering a little bit why you use `if [[ -z "${STAMP//Running online/}" ]]; then` ?

fbmei:
I use it to prevent the overwrite the old stamp which i assume would happen if the database is already stamped to an older version to not interfere with the upgrade. The command `edumfa-manage db current -d /usr/local/lib/edumfa/migrations 2>/dev/null` returns always `Running online` so it gets stripped away and if the rest of the string is empty the DB is not already stamped.

fritterhoff (approved):

PR #456, opened 2024-12-01
fix: url escaping of username and password fields for sql connections

fbmei:
The SQLIdResolver doesn't escape username and password fields when creating the DB connection string in the _create_connect_string method. This leads to connection errors when using special chars like ':/@' in them. I added url escaping using the urllib.parse.quote method and created a test for the _create_connect_string method in the SQLResolverTestCase in test_lib_resolver.py.


fritterhoff:
Mhm. Does it make sense to use `URL.create(` and rewrite the segment using a URL object?

fbmei:
Ok, I changed it to URL.create(). Thanks for the hint; I didn't know SQLAlchemy had a method for that. I added the option to use the conParams, which were previously unused. To define additional parameters in the web UI, they need to be passed as a JSON string in the form of { "option": "value" }. Is this okay like that? I changed the code in the resolver to use the URL object directly rather than the string because it's shown that way in the SQLAlchemy docs, and I didn't find a way to export the string from the URL object without the password obfuscation. I also removed the test I wrote earlier because it's not needed anymore if SQLAlchemy does all the heavy lifting. I changed the tests that use SQLite databases from the testdata directory too because they were using the server parameter of the resolver to store a part of the database path. This isn't needed anymore because it can be stored in the database parameter now, where it belongs.

Changes made:

edumfa/lib/auditmodules/sqlaudit.py:
- removed usage of censor_connect_string
- refactored connect_string to connect_url

edumfa/lib/resolvers/SQLIdResolver.py:
- added support for json encoded additional connection parameters
- implemented _create_connect_string with sqlalchemy.URL.create()
- refactored _create_connect_string to _create_connect_url
- refactored connect_string to connect_url
- refactored self.connect_string to self.connect_url
- removed usage of censor_connect_string

edumfa/lib/utils/__init__.py:
- removed censor_connect_string helper method

tests/test_lib_utils.py
- removed test for censor_connect_string

tests/test_lib_resolver.py:
- removed test for create_connect_string
- removed use of servername for db path

lib/monitoringmodules/sqlstats.py
- removed usage of censor_connect_string
    
tests/base.py
tests/test_api_register.py
tests/test_api_users.py
tests/test_lib_eventhandler_usernotification.py
tests/test_lib_recovery.py
tests/test_lib_user.py
tests/test_lib_usercache.py
- removed use of server parameter for the db path

tests/base.py
```diff
@@ -159,8 +159,8 @@ def setUp_sqlite_resolver_realm(self, sqlite_file, realm):
         parameters = {'resolver': "sqlite_resolver",
                       "type": "sqlresolver",
                       'Driver': 'sqlite',
-                      'Server': '/tests/testdata/',
-                      'Database': sqlite_file,
+                      'Server': '',
+                      'Database': 'tests/testdata/' + sqlite_file,
```
fritterhoff, line 163: Did not test it right now but are be backward compatible?

fbmei: With older versions of SQLAlchemy or with older testcode and are you referring to the usage of the Database parameter instead of the Server parameter or the string concatenation in this special case ?

fritterhoff: With older configurations. So does it break existing configs that use the separated definitions?

fbmei: Ah ok yes it would break if someone defined an SQLResolver with an SQLite DB like this. I assumed no one outside of the tests would do something like this because it's not documented anywhere and the WebUI has clear definitions for the fields. Is there an update mechanism which i could use to parse it and convert it if needed ? Like the db upgrade functionality. Otherwise i would try to add logic to the _create_connection_url method to detect and correct this.

fritterhoff: Maybe a SQL Migration could work. Let me discuss it on Friday with the remaining team.

fbmei: Ok thanks.

fritterhoff: We are quite unsure if this config is really not used by anyone. Would prefer to have an migration or at lease some sort of backward compatibility.

PR #465, opened 2024-12-04
feat: disable timeshift

j-hoff:
add option to TOTP token config to globally disable use of timeshift values for TOTP tokens, token config can be overwritten per token by adding a tokenInfo attribute


fritterhoff:
@j-hoff whats the status of this PR? We are now running into similar issues and would consider to try this workaround :)

j-hoff:
@fritterhoff So far it's already usable and can be enabled globally in the token config or per token via tokeninfo attribute "useTimeShift".
What's missing is the option to modify the tokeninfo attribute via GUI. I can probably continue working on this soon.

fritterhoff:
Since we are running into similar (strange) issues I think it would make sense to merge this without UI option and only provide a API method

j-hoff:
OK, feel free to test the current state, add anything. and/or merge.
It's still missing some doc additions and test case.

j-hoff:
@fritterhoff Fixed some issues with this and added the possibility to set the useTimeShift-tokeninfo attribute on enrollment and edit it on the token details view in the UI.
If this is set to True or False it will override the global default (checkbox in TOTP token config).

fritterhoff (approved):

PR #474, opened 2024-12-10
fix: add check to exec user scripts only if scripts are available

### LINKED ISSUE #472: docker compose fails with "/opt/edumfa/user-scripts/*.sh: No such file or directory"
### Top-level intent

Follow-up of #360. Running the docker compose instructions leads to the eduMFA container erroring out with:

```
edumfa-1  |                _       __  __ ______      
edumfa-1  |               | |     |  \/  |  ____/\    
edumfa-1  |        ___  __| |_   _| \  / | |__ /  \   
edumfa-1  |       / _ \/ _` | | | | |\/| |  __/ /\ \  
edumfa-1  |      |  __/ (_| | |_| | |  | | | / ____ \ 
edumfa-1  |       \___|\__,_|\__,_|_|  |_|_|/_/    \_\       v2.5.0
edumfa-1  |     
edumfa-1  | Admin admin was registered successfully.
edumfa-1  | Executing User-Scripts
edumfa-1  | bash: /opt/edumfa/user-scripts/*.sh: No such file or directory
edumfa-1 exited with code 127
```

### Steps to reproduce

1. Grab the `docker-compose.yaml` from [the docs](https://edumfa.readthedocs.io/en/v2.3.0/installation/docker.html#docker-compose) or the [latest docs](https://edumfa.readthedocs.io/en/latest/installation/docker.html#docker-compose)
2. Create a configuration file and adapt the path to it in the `docker-compose.yaml` 
3. Run `docker compose up`

### Expected outcome

The container should start successfully.

### Actual outcome

Scripts are not found, see above.

fbmei:
fixes issue #472

### DIFF
--- a/deploy/docker-setup.sh
+++ b/deploy/docker-setup.sh
@@ -32,9 +32,11 @@ edumfa-manage db upgrade -d /usr/local/lib/edumfa/migrations
 echo "Creating Admin"
 edumfa-manage admin add $EDUMFA_ADMIN_USER -p $EDUMFA_ADMIN_PASS
 
-echo "Executing User-Scripts"
 # Execute user scripts
-bash /opt/edumfa/user-scripts/*.sh
+if [ "$(ls -A /opt/edumfa/user-scripts/)" ]; then
+    echo "Executing User-Scripts"
+    bash /opt/edumfa/user-scripts/*.sh
+fi
 
 echo "Starting Server"
 gunicorn --bind 0.0.0.0:8000 --workers 4 app
--- a/deploy/docker-setup.sh
+++ b/deploy/docker-setup.sh
@@ -33,7 +33,7 @@ echo "Creating Admin"
 edumfa-manage admin add $EDUMFA_ADMIN_USER -p $EDUMFA_ADMIN_PASS
 
 # Execute user scripts
-if [ "$(ls -A /opt/edumfa/user-scripts/)" ]; then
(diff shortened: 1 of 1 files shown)


deploy/docker-setup.sh
```diff
@@ -32,9 +32,11 @@ edumfa-manage db upgrade -d /usr/local/lib/edumfa/migrations
 echo "Creating Admin"
 edumfa-manage admin add $EDUMFA_ADMIN_USER -p $EDUMFA_ADMIN_PASS
 
-echo "Executing User-Scripts"
 # Execute user scripts
-bash /opt/edumfa/user-scripts/*.sh
+if [ "$(ls -A /opt/edumfa/user-scripts/)" ]; then
```
fritterhoff: Mhm. I considered using: 
`if compgen -G "/opt/edumfa/user-scripts/*.sh" > /dev/null; then `

fritterhoff (changes requested):

deploy/docker-setup.sh
fbmei: Yes this works thanks. I updated the branch. Are there any advantages from using compgen or is it because it's a builtin ?

fritterhoff: compgen allows to find shell scripts. So in case there are any other files like a README.md or something like that it prop. would break in your case...

fbmei: Ah yes thats true. Didn't think someone would place anything else there.

fritterhoff (approved):

PR #494, opened 2025-01-07
fix: use format string for last auth timestamp

j-hoff:
In the rare cases where the microseconds of a last_auth timestamp are 0, they were omitted:
```
>>> print(datetime(2025,6,1,12,0,0,0,tzinfo=tzlocal()))
2025-06-01 12:00:00+02:00
>>> print(datetime(2025,6,1,12,0,0,1,tzinfo=tzlocal()))
2025-06-01 12:00:00.000001+02:00
```
This change uses the format string '%Y-%m-%d %H:%M:%S.%f%:z'
 when writing the 'last_auth' time to tokeninfo,
 so nothing changes for timestamps with microseconds!=0, but we get a consistent format:
```
print(datetime(2025,6,1,12,0,0,0).strftime('%Y-%m-%d %H:%M:%S.%f%:z'))
2025-06-01 12:00:00.000000+02:00
```

### DIFF
--- a/edumfa/lib/policydecorators.py
+++ b/edumfa/lib/policydecorators.py
@@ -32,6 +32,7 @@ from edumfa.lib.policy import Match
 from edumfa.lib.error import PolicyError, eduMFAError
 import functools
 from edumfa.lib.policy import ACTION, SCOPE, ACTIONVALUE, LOGINMODE
+from edumfa.lib.tokenclass import AUTH_DATE_FORMAT
 from edumfa.lib.user import User
 from edumfa.lib.utils import parse_timelimit, parse_timedelta, split_pin_pass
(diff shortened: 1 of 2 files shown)


fritterhoff:
I am wondering if we have similar problems somewhere else?

j-hoff:
Another place I found, where timestamps are written is the token handler. There we already have fixed formats.
(`{now} and {current_time} in tokeninfo get replaced by a timestamp that has almost the same format as last_auth, but without the ':' in the UTC offset. 
validity_perios_start/_end has another format without seconds.)

It might be a good idea to align those to a uniform format in the future, but will break user code, which relies on the format.

fritterhoff:
Ok. Then I would merge this PR and streamline the format in one of the next versions.

fritterhoff (approved):

PR #508, opened 2025-01-14
feat: add conditional formatting of auth timestamp

fritterhoff:


j-hoff:
created a PR which should work for all supported python versions now: #510

PR #511, opened 2025-01-15
feat: start adding audience check

### LINKED ISSUE #497: Trusted JWTs miss `aud` check
Trusted JWT should be used to authenticate a user in eduMFA to make applications able to use the API without storing master credentials on the server.

The current implementation would mean that all tokens from all clients from a identity provider where able to impersonate with the configured user since no audience is checked.

The implementation should be changed.
eduMFA should be configurable with the following properties:  
- Trusted authority
- JWKS Urls to validate the certificate
- Trusted audiences
- user-claim

Now you can validate the token and extract the user from it to login as this user so that you can only login with your own account.

fritterhoff:


fritterhoff:
Fixes #497

PR #512, opened 2025-01-15
ci: add docs in ci

fritterhoff:


fritterhoff:
@Luc1412 a really hacky workaround... prop we should try to get rid of the crypto stuff...

PR #520, opened 2025-01-20
registration codes can be explicitly set via otpkey parameter

pmainz:
Registration codes (tokentype registration) are an easy way of initially rolling out a token. Right now these codes can only be generated by eduMFA. Since rollout processes often involve automatic token creation via API and are embedded in more complex processes, one might want to generate the otp-value in advance and then explicitly set it in eduMFA.

This PR removes the restriction that registration codes always set genkey to 1 and so makes it possible to set the otp via otpkey parameter as in many other token types.


j-hoff (approved):
looks fine

PR #531, opened 2025-02-02
fix session handling

fritterhoff:
Right now the sessions are not cleanup correctly after each response. The finalizer is only called the first time and afterwards not registered again.


j-hoff (approved):
this seems consistent

PR #533, opened 2025-02-03
feat: add option to define error behavior on orphan

### LINKED ISSUE #525: copied: privacyidea-token-janitor mistakenly unassigns (many) tokens on LDAP errors
Hello,

I have noticed that eduMFA also contains the same (incorrect) code.
Should the presented change also be migrated to eduMFA?

https://github.com/privacyidea/privacyidea/issues/4204

fritterhoff:
fixes #525

### DIFF
--- a/tools/edumfa-token-janitor
+++ b/tools/edumfa-token-janitor
@@ -276,7 +276,7 @@ def build_tokenvalue_filter(key,
 
 def _get_tokenlist(last_auth, assigned, active, tokeninfo_key,
                    tokeninfo_value_filter, tokenattribute, tokenattribute_filter,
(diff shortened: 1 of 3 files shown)


tools/edumfa-token-janitor
```diff
@@ -473,6 +473,8 @@ def export_user_data(token_list, attributes=None):
 @click.option('--active', help='True|False|None')
 @click.option('--orphaned',
               help='Whether the token is an orphaned token. Set to 1')
+@click.option('--orphan_exception_value', default=True, type=bool,
```
pmainz: Would it make sense to set the 'False' as the default, just to prevent more people from accidentally deleting tokens?  Otherwise I think we should add a note to the help text here... like 'If not set to False tokens are handled as orphaned in case the userstore is temporarily not available.'

pmainz (approved):

PR #552, opened 2025-02-27
Support custom paths in config for ubuntu install scripts

### LINKED ISSUE #460: Update not possible because of different path in EDUMFA_ENCFILE
An update of edumfa is not possible if a different path is specified in EDUMFA_ENCFILE.
I think the bug is in 

> deploy/ubuntu-server/edumfa-apache2.postinst line 44

### Steps to reproduce

1. edit /etc/edumfa/edumfa.cfg: EDUMFA_ENCFILE = '/etc/edumfa/secrets/enckey'
2. move enckey to this path
3. run update

### Expected outcome

successful update

### Actual outcome

failed update

### Configuration

* **eduMFA version**: 2.3.0 
* **Installation method**: from Ubuntu packages
* **Python version**: 3.12.3
* **Operating system**: ubuntu 24.04
* **Webserver**: apache
* **Token database**: MySQL

### Log file

``` log
chmod: cannot access '/etc/edumfa/enckey': No such file or directory
dpkg: Fehler beim Bearbeiten des Paketes edumfa-apache2 (--configure):

```

Luc1412:
This PR fixes #460 by supporting custom paths set in the config while using the ubuntu apache or nginx package.

These changes aren't tested yet.


deploy/ubuntu-server/edumfa-nginx.postinst
```diff
@@ -58,10 +110,11 @@ create_certificate() {
     openssl x509 -req -days $CERTDAYS -in server.csr -signkey server.key -out server.crt
     cat server.crt server.key > server.pem
     rm -f server.crt server.key
-    chown edumfa server.pem
+    chown "$USERNAME" server.pem
     chmod 400 server.pem
   fi
   
+  # Rest of the function remains unchanged
```
aleyna72072: this will be a confusing comment in the future :smile:

aleyna72072:
Tested the following:
- installed this branch (2.7.0)
- changed edumfa.cfg like this:
```
EDUMFA_LOGFILE = '/var/log/edumfa_luc/edumfa.log'
EDUMFA_ENCFILE = '/etc/edumfa_luc/enckey'
EDUMFA_AUDIT_KEY_PRIVATE = '/etc/edumfa_luc/private.pem'
EDUMFA_AUDIT_KEY_PUBLIC = '/etc/edumfa_luc/public.pem'
```
- moved the files/directory likewise
- installed main branch with this patch

Everything seems to work fine.

aleyna72072:
Remaining tests:
1. normal upgrade without any non-standard settings
2. what happens when the config file is syntactically different but still valid

Anything you'd like me to test, too?

aleyna72072:
> normal upgrade without any non-standard settings

seems to work

aleyna72072:
> what happens when the config file is syntactically different but still valid

`EDUMFA_ENCFILE="/etc/edumfa_luc/enckey"` instead of `EDUMFA_ENCFILE = '/etc/edumfa_luc/enckey'` works

What does not work:
- Obviously using any "code" to configure the value.
  + i.e. something like `f"{os.getcwd()}/enckey`
  + I don't know if anybody does this though.
- If the key is defined multiple times the server works but the upgrade fails, i.e.:
```
EDUMFA_ENCFILE = 'quatsch'
EDUMFA_ENCFILE = '/etc/edumfa_luc/enckey'
```

I could maybe see the last one happening on accident and not being noticed? At least the error message on upgrade is clear enough for somebody to figure out what is going on:
```
chmod: cannot access 'quatsch'$'\n''/etc/edumfa_luc/enckey': No such file or directory
dpkg: error processing package edumfa-apache2 (--configure):
 installed edumfa-apache2 package post-installation script subprocess returned error exit status 1
Errors were encountered while processing:
 edumfa-apache2
needrestart is being skipped since dpkg has failed
E: Sub-process /usr/bin/dpkg returned an error code (1)
```
I can only see this being fixed if you'd parse the cfg Python-native but that's probably introducing more overhead than it's worth?
If this is getting merged, the first note on https://edumfa.readthedocs.io/en/v2.8.0/installation/system/inifile.html needs to be updated though, as
> The config file is parsed as python code, so you can use variables to set the path and you need to take care of the indentation.

would no longer be supported in Ubuntu packages.

Luc1412:
> Anything you'd like me to test, too?

I think your tests look pretty good. All necessary should be covered.

> Obviously using any "code" to configure the value.

I think that should be fine. It's probably pretty niche.

> If the key is defined multiple times the server works but the upgrade fails, i.e.:

I'd say we don't need to account for misconfigurations

> I can only see this being fixed if you'd parse the cfg Python-native but that's probably introducing more overhead than it's worth?

Probably. If there is demand, a feature request can be opened to discuss this further.

> If this is getting merged, the first note on https://edumfa.readthedocs.io/en/v2.8.0/installation/system/inifile.html needs to be updated though

Thanks for lmk. Added this to the PR.

@aleyna72072 If you feel like everything looks fine, please add an approving review.

aleyna72072:
I just noticed I only tested the -apache2 package, not the -nginx one. I also did not test the new commit.  
I'll ask the company I work at if I'll get the time to also test these things.

aleyna72072 (approved):
Tested -nginx -apache2, both work without issues.

fritterhoff (approved):

PR #607, opened 2025-04-16
fix: make challenge cleanup & client applications more reslient

fritterhoff:
right now there are some minor corner cases where the session handling may crash under unclear corner cases.

Some of them are not that urgent and can simply get ignored by wrapping them into a try catch

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,7 +1,0 @@
-{
-    "python.testing.pytestArgs": [
-        "tests"
-    ],
-    "python.testing.unittestEnabled": false,
-    "python.testing.pytestEnabled": true
-}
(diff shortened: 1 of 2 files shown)


Luc1412 (commented):
Is it possible to not use broad exception catching and rather define specific exceptions that should be catched?

.vscode/settings.json
Luc1412, line 1: This file shouldn't be part of the PR

Luc1412 (changes requested):

.vscode/settings.json
fritterhoff: Sry!

Luc1412 (approved):

Issue #670, opened 2025-06-12
ldapresolver change does not take effect immediatly


aleyna72072:
### Top-level intent
Changing the config of an ldapresolver.

### Steps to reproduce

1. Set up an ldapresolver and save it (e.g. with base-dn pointing to an empty ou).
2. Modify and save it (e.g. point it to the correct base-dn).
3. Check if changes are visible (e.g. see if users appear in user list).
This also means that after a modification, the old config is used and users are able to log in via it.

### Expected outcome
Changes are visible instantly.

### Actual outcome
The user list is as before the modification. 

### Configuration

* **eduMFA version**: 2.8.0
* **Installation method**: Ubuntu packages
* **Python version**: 3.12.3
* **Operating system**: Ubuntu 24.04
* **Webserver**: Apache2
* **Token database**: MySQL

aleyna72072:
Reverting d07e84075756d639d6977d4c50ae8d43789cd8c6 fixes it. Digging further..

fritterhoff:
Hi! Jep. We are aware of this issue. Right now we only observe the problem in case of multiple instances. These would require some sort of syncronisation

aleyna72072:
Problem: All nodes/processes have to be restarted when changing an LDAP resolver, therefore just destroying connections in one won't work.
Possible solution: Popup saying e.g. "Restart eduMFA for these changes to have effect." when saving an LDAP resolver.

PR #671, opened 2025-06-12
fix: delete existing connections when saving a resolver

aleyna72072:
When modifying e.g. an LDAP connection, it currently re-uses the old configuration. Therefore delete the resolver object so it gets re-created on the next request, avoiding stale data.
I tested with the following resolvers:

- [x] LDAP
- [x] passwd
- [x] SQL
- [x] LDAP+passwd

I did not test with a SCIM or HTTP resolver.


aleyna72072:
https://github.com/eduMFA/eduMFA/issues/670#issuecomment-2969517892

PR #676, opened 2025-06-17
feat: restart request dialog on ldapresolver change

### LINKED ISSUE #670: ldapresolver change does not take effect immediatly

aleyna72072:
When modifying e.g. an LDAP connection, it currently re-uses the old configuration. This adds a popup informing the user to restart their eduMFA:
image

Without additional complexity (e.g. requesting the current list of resolvers), it is unknown whether the `resolvername` already exists (i.e. whether we are editing an existing resolver or a new one). This is especially so since users able to enter any name into the editing mask even when pressing "New ldapresolver", potentially modifying an already existing resolver. To keep things simple the pop-up shows every time an LDAPresolver is saved.
fixes #670

### DIFF
--- a/edumfa/static/components/login/controllers/loginControllers.js
+++ b/edumfa/static/components/login/controllers/loginControllers.js
@@ -640,6 +640,10 @@ angular.module("eduMfaApp")
         $scope.$broadcast("piReload");
     };
 
+    $scope.ldapRestartDialog = function() {
+        $("#dialogLdapRestart").modal("show");
+    };
+
 }]);
 
 angular.module("eduMfaApp")
--- a/edumfa/static/components/config/controllers/configControllers.js
+++ b/edumfa/static/components/config/controllers/configControllers.js
@@ -1079,6 +1079,9 @@ myApp.controller("LdapResolverController", ["$scope", "ConfigFactory", "$state",
         ConfigFactory.setResolver($scope.resolvername, $scope.params, function (data) {
             $scope.set_result = data.result.value;
             $scope.getResolvers();
+            // As we don't know whether resolvername already exists, just show
+            // the pop-up each time to keep things simple.
+            $scope.ldapRestartDialog();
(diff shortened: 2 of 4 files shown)


aleyna72072:
@fritterhoff What do you think? :slightly_smiling_face:

aleyna72072:
FYI: I did try to be more clever at https://github.com/aleyna72072/eduMFA/tree/feature/modifyldaprestartpopup but since you can just enter whatever for the resolvername, I think this is the better solution even if annoying.

edumfa/static/components/login/controllers/loginControllers.js
```diff
@@ -640,6 +640,10 @@ angular.module("eduMfaApp")
         $scope.$broadcast("piReload");
     };
 
+    $scope.ldapRestartDialog = function() {
+        $("#dialogLdapRestart").modal();
```
aleyna72072: TODO: Maybe this should be model("show") instead? If the button is pressed two times extremely quickly.

fritterhoff: Think for now that should be fine....

fritterhoff (approved):

PR #680, opened 2025-06-20
fix: move edumfa venv install prefix to rules file

aleyna72072:
When building your own package as described [in the docs](https://edumfa.readthedocs.io/en/v2.8.0/installation/ubuntu.html#building-your-own-packages), you'd get a wrong path (and broken packages) as the default prefix is /opt/venvs.  
This patch sets the environment variable as described in dh_virtualenv's [documentation](https://dh-virtualenv.readthedocs.io/en/latest/usage.html?highlight=venvs).
Also "-Sa" is not a valid option for dpkg source, error: `dpkg-source: warning: -Sa is not a valid option for Dpkg::Source::Package::V3::Quilt`. I assume this is a misspelling of -sa/-sA (see manpage), so I removed it.


aleyna72072:
@Luc1412 I think this might also be your area of expertise? :slightly_smiling_face:

Luc1412 (approved):
Good change 👍 Removing flag shouldn't be an issue, when it worked fine before

PR #685, opened 2025-06-24
chore: migrate timestamps to timezone-aware UTC datetimes

Luc1412:
## Summary
This PR migrates the project from naive `datetime.utcnow()` to timezone-aware UTC timestamps to prevent ambiguity and ensure consistent datetime handling across the application, database, and tests.

## Key Changes
- **Utilities**: Added `utc_now()` in `edumfa/lib/utils/__init__.py` as the standard way to get aware UTC timestamps.
- **ORM/Database**: Introduced `UTCDateTime` in `edumfa/models.py` to ensure DB values normalize to UTC and return as timezone-aware objects.
- **Migration**: Added Alembic migration `migrations/versions/e0b3e8c1f9a2_.py` to update timestamp columns.
- **Codebase Migration**: Replaced all `datetime.utcnow()` calls with `utc_now()`, including background scripts like `tools/edumfa-expired-users`.
- **Tests**: Updated the test suite to use and expect timezone-aware UTC datetimes, while maintaining compatibility for audit signature verification.


aleyna72072:
Rebased to see how it looks with the new unittests. Looking forward to see this merged, the warnings are annoying.  
Since my matrix messages are unreadable: the reason for the draft status was this is potentially a breaking change?

Luc1412:
> Since my matrix messages are unreadable: the reason for the draft status was this is potentially a breaking change?

Yeah + it's untested.

aleyna72072:
I would consider this a breaking change, as the timestamps e.g. in the Audit table could change their meaning.

edumfa/lib/utils/__init__.py
aleyna72072, line 99: Marker for me to figure out if this is still correct.

aleyna72072, line 1397: Note to self: this is correct, you probably want to be able to control the time your users see by choosing the server's TZ.

PR #691, opened 2025-06-30
fix: ldap resolvers with equal conf except binddn

### LINKED ISSUE #690: cache confusion on LDAP-Resolvers with equal configuration but the binddn
### Top-level intent
Create two equal resolver, which deliver different results depending on the bind_dn user.

### Steps to reproduce
1. Create two LDAP resolvers pointing at the same LDAP with different bind_dn's.
2. Make sure it sends different results depending on their bind_dn.
3. Try to login with each.

### Expected outcome
Both logins work.

### Actual outcome
Only one login works.

### Configuration

* **eduMFA version**: 2.8.0
* **Installation method**: PyPI
* **Python version**: 3.12.3
* **Operating system**: Ubuntu Noble
* **Webserver**: Apache2
* **Token database**: MySQL

### Resolution?
1. getUserId is cached: https://github.com/eduMFA/eduMFA/blob/v2.8.0/edumfa/lib/resolvers/LDAPIdResolver.py#L625
2. the cache looks for a generated resolver id: https://github.com/eduMFA/eduMFA/blob/v2.8.0/edumfa/lib/resolvers/LDAPIdResolver.py#L218
3. the resolver id does not contain the binddn: https://github.com/eduMFA/eduMFA/blob/v2.8.0/edumfa/lib/resolvers/LDAPIdResolver.py#L746

The easiest fix is to just use the binddn for the ID, too.

aleyna72072:
Currently, the cache ID does not contain the bind dn. This means that for configurations where the bind dn has an effect on the ldap content, it can lead to confusions in the cache when there are otherwise two equal resolvers.

- [x]  Unittest: update existing and write new one for this specific case
- [x] testing/make sure this does not break stuff
- [x]  The sql resolver adds a prefix, maybe use one here too?
- [x] Why the NULL byte? https://github.com/eduMFA/eduMFA/blob/main/edumfa/lib/resolvers/SQLIdResolver.py#L417

### DIFF
--- a/edumfa/lib/resolvers/LDAPIdResolver.py
+++ b/edumfa/lib/resolvers/LDAPIdResolver.py
@@ -752,11 +752,12 @@ class IdResolver (UserIdResolver):
         :return: the id of the resolver
         :rtype: str
         """
-        s = "{0!s}{1!s}{2!s}{3!s}".format(self.uri, self.basedn,
+        s = "{0!s}{1!s}{2!s}{3!s}{4!s}".format(self.uri, self.basedn,
+                                          self.binddn,
                                           self.searchfilter,
                                           sorted(self.userinfo.items(), key=itemgetter(0)))
         r = binascii.hexlify(hashlib.sha1(s.encode("utf-8")).digest())  # nosec B324 # hash used as unique identifier
-        return r.decode('utf8')
+        return "ldap." + r.decode('utf8')
 
     @staticmethod
     def getResolverClassType():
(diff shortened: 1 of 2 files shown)


aleyna72072:
I'll come back to this hopefully next month.

aleyna72072:
https://github.com/eduMFA/eduMFA/commit/9432e376fc055901fadfbf9d00ffb05e58a2cc92
NULL byte seems just to be used as an easy way to concat each value. Ignore.

Luc1412 (approved):

PR #694, opened 2025-07-03
feat(firebase): add configurable timeouts for firebase requests

jan-peterfriebe:
…w polling if firebase is down with Shibboleth frontend proxy.

Hello there,
At the University of Bamberg, we use [Shibboleth](https://en.wikipedia.org/wiki/Shibboleth_(software)) as our authentication agent. It integrates well with edumfa. We were concerned about reliability if the internet went down or Firebase isn't reachable anymore. We blocked all Google Firebase API servers via our firewall to test the polling method. Each time, after 60 seconds, Shibboleth would kick you out because it thought the service had gone down, while it was processing in the background for a long time.

#### Added Try-Except Statement

We added a try-except statement around the POST request to the Firebase server. So if the server isn't reachable, it would not crash.
#### Added Timeout

We added a timeout parameter to the Firebase request so that google.auth + request wouldn't try forever to connect to Firebase. We also added configurable parameters to edumfa.cfg to change the timeout parameters.

There was already a similar attempt #338, but the issue isn't fixed yet #275.

### DIFF
--- a/deploy/edumfa/edumfa.cfg
+++ b/deploy/edumfa/edumfa.cfg
@@ -24,3 +24,13 @@ EDUMFA_AUDIT_POOL_SIZE = 20
 
 EDUMFA_LOGFILE = '/var/log/edumfa/edumfa.log'
 EDUMFA_LOGLEVEL = logging.INFO
+
+
+# Timeout duration (in seconds) for establishing a connection to the Firebase server.
+# If the server does not respond within this time, the connection attempt is aborted.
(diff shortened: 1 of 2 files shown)


fritterhoff:
Hi! Thanks for the changes. Could you please fix the issues that are the breaking the tests.
Thanks

fritterhoff (changes requested):
Code must be fixed to get tests working again.

fritterhoff (approved):

PR #704, opened 2025-07-18
docs: add hardware requirements rec and some fixes

aleyna72072:
This adds a small section on the overview so people have _some_ idea where to start for an eduMFA VM. In the long run maybe some real numbers from deployments could be added (requests/second, load, specs, etc)?  

grafik

Also fixes some mistakes in the documentation which I've been collecting.


aleyna72072:
@Luc1412  Maybe this is something you could take a look at?

Luc1412 (approved):

PR #724, opened 2025-08-08
feat: pin docker digest and use specific python version

fritterhoff:
use explicit version to enable renovate to update the docker image more often and get a up-to-date version

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,12 +1,12 @@
 # Build stage
-FROM python:3.13-slim-bookworm AS builder
+FROM python:3.13.6-slim@sha256:6f79e7a10bb7d0b0a50534a70ebc78823f941fba26143ecd7e6c5dca9d7d7e8a AS builder
 WORKDIR /tmp
 COPY . .
 RUN pip install --no-cache-dir build && \
     python -m build --sdist --wheel --outdir dist/
 
 # Final stage
-FROM python:3.13-slim-bookworm
+FROM python:3.13.6-slim@sha256:6f79e7a10bb7d0b0a50534a70ebc78823f941fba26143ecd7e6c5dca9d7d7e8a
 
 # Install system dependencies
 RUN apt-get update && \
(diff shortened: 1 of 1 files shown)


aleyna72072:
[Since this is changing away from the bookworm tag](https://hub.docker.com/_/python#pythonversion):
> Some of these tags may have names like bookworm or bullseye in them. These are the suite code names for releases of [Debian⁠](https://wiki.debian.org/DebianReleases) and indicate which release the image is based on. If your image needs to install any additional packages beyond what comes with the image, you'll likely want to specify one of these explicitly to minimize breakage when there are new releases of Debian.

We do install the following packages:
> apt-get install -y curl libpq-dev gcc

Using the ("highly recommended") non-slim-image, all of these would already be installed, but the image is apparently 10x bigger..
```
$ docker image ls | grep py
python                                                                     3.13.6-slim                   52a3073fe514   35 hours ago    120MB
python                                                                     3.13.6                        3e7f48ebe9d8   35 hours ago    1.02GB
```

So maybe slim-bookworm (or starting tomorrow soon slim-trixie?) is a better idea to not be confused by breaking changes between major versions?

fritterhoff (commented):
fix: stick to bookworm

aleyna72072 (approved):
Da wir schon eine Build-Stage haben, könnten wir in Zukunft mglw. auf [distroless](https://github.com/GoogleContainerTools/distroless/tree/main/python3)/[chainguard](https://github.com/chainguard-images/images/tree/main/images/python) wechseln?

Luc1412 (changes requested):
I'd say for better tracking changes down, we should add a PR description to provide some reasoning behind the change, because it might not be clear to all

fritterhoff:
> Da wir schon eine Build-Stage haben, könnten wir in Zukunft mglw. auf [distroless](https://github.com/GoogleContainerTools/distroless/tree/main/python3)/[chainguard](https://github.com/chainguard-images/images/tree/main/images/python) wechseln?

would preferr to stick to the official images

fritterhoff:
> I'd say for better tracking changes down, we should add a PR description to provide some reasoning behind the change, because it might not be clear to all

updated the description

PR #727, opened 2025-08-08
feat: store webauthn `credential_id` unencrypted

fritterhoff:
Store public WebAuthn credential IDs without encryption to remove unnecessary crypto work during authentication. It also adds a migration for existing tokens, keeps encrypted legacy entries usable when migration cannot decrypt them, and support downgrades by re-encrypting credential IDs. This workflow also ensures, that the instance stays usable even if the migration is slow and is e.g. running in background.

### DIFF
--- a/migrations/versions/1b2262ddbf6b_.py
+++ b/migrations/versions/1b2262ddbf6b_.py
@@ -1,0 +1,40 @@
+"""Decrypt WebAuthn tokens. 
+
+Before this version, WebAuthn tokens were stored in an encrypted format. This migration will decrypt the tokens and store them in a plain format. 
+Decrypting the credential id is not really necessary, as it is not sensitive information.
+
+Revision ID: 1b2262ddbf6b
+Revises: a844dc0c3fb9
+Create Date: 2025-08-23 08:05:06.963794
+
+"""
+
+# revision identifiers, used by Alembic.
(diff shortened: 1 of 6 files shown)


migrations/versions/1b2262ddbf6b_.py
```diff
@@ -0,0 +1,158 @@
...
+    print("Storing the credential id of WebAuthn tokens unencrypted")
+    # Fail early with a meaningful error, if the security module is not ready.
+    get_hsm()
+    migrated, failed = decrypt_credential_ids(op.get_bind())
+    print(f" -> Decrypted the credential id of {migrated} WebAuthn token(s).")
+    if failed:
+        print(
+            f" -> WARNING: The credential id of {failed} WebAuthn token(s) could not be "
+            "decrypted. These tokens keep their encrypted credential id and continue "
+            "to work.",
```
aleyna72072, line 149:
```suggestion
            "decrypted. These tokens keep their encrypted credential id and were "
            "not changed.",
```

aleyna72072 (approved):
LGTM.  
Just out of curiosity: are broken WebAuthN tokens common? How do they come to exist?

fritterhoff:
Hopefully there are none 😁 that is more a sanity check added as precaution.

aleyna72072 (approved):
LGTM

PR #734, opened 2025-08-18
fix: define sequence start and minvalue explicitly

### LINKED ISSUE #733: Sequences are not properly created when using MSSQL
### Top-level intent

Initialize the database structure on a MSSQL database server

### Steps to reproduce

1. Install eduMFA
2. Create databes with `edumfa-manage create_tables`

### Expected outcome

The created sequences have a start value and min value of `1`.

### Actual outcome

The created sequences have a start value and min value of `-2 ^ 63`.

### Configuration

* **eduMFA version**: 2.8.0
* **Installation method**: PyPI
* **Python version**: 3.12
* **Operating system**: AlmaLinux 10
* **Webserver**: httpd
* **Token database**: MSSQL

Thaoden:
This addresses #733 
The start value and min value of all sequences have been explicitly set to `1` to address a change in SQLAlchemy that affects MSSQL databases.

### DIFF
--- a/edumfa/models.py
+++ b/edumfa/models.py
@@ -162,7 +162,7 @@ class Token(MethodsMixin, db.Model):
     """
     __tablename__ = 'token'
     __table_args__ = {'mysql_row_format': 'DYNAMIC'}
-    id = db.Column(db.Integer, Sequence("token_seq"),
+    id = db.Column(db.Integer, Sequence("token_seq", start=1, minvalue=1),
                    primary_key=True,
                    nullable=False)
     description = db.Column(db.Unicode(80), default='')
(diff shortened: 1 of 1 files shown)


fritterhoff:
Hi! We discuseed that internally and are a bit wondering if a migration would be required for existing deployments. Will check that in some test deployments and update you.

fritterhoff:
So we discussed this longer last week. We would perfer a solution that only has impact on the mssql part.

In theory it should be possible to build sth like this and manipulate the sql string...

```
@compiles(CreateSequence, 'mssql')
def fix_start_offset(element, compiler, **kw):  # pragma: no cover
    text = compiler.visit_create_sequence(element, **kw)
    return text

```

Thaoden:
While I understand the desire to only impact the providers having an issue right now, I admittedly don't understand the separation here. For MSSQL, this fixes the bug, while for other providers, it merely makes the implicit default values explicit and hence doesn't break existing scenarios - which would be a benefit, in case the default values ever changed.

My own mastery of Python isn't good enough to implement the suggested SQL string manipluation.

Woo6vop7:
Any news or plans on this?

aleyna72072:
Closing since eduMFA now has explicitly supported databases (#1163) and MSSQL is not one of them.

PR #736, opened 2025-08-22
fix: only show attributes if change is allowed

cbergmann:
The attribute list is always shown to admins even if they are not allowed to set/delete them.
This is because the template looks for existance of the key and [policy.py](https://github.com/eduMFA/eduMFA/blob/main/edumfa/lib/policy.py#L2945) always returns a object with the key set. If the user has no rights the keys are also set but the list ist empty. This change updates the template to reflect this.

### DIFF
--- a/edumfa/static/components/user/views/user.details.html
+++ b/edumfa/static/components/user/views/user.details.html
@@ -75,7 +75,7 @@
     </div>
     -->
 
-    <div ng-show="allowed_custom_attributes.hasOwnProperty('set') || allowed_custom_attributes.hasOwnProperty('delete')">
+    <div ng-show="allowed_custom_attributes['set'].length > 0 || allowed_custom_attributes['delete'].length > 0">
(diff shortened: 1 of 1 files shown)


edumfa/static/components/user/views/user.details.html
```diff
@@ -75,7 +75,7 @@ <h2>
     </div>
     -->
 
-    <div ng-show="allowed_custom_attributes.hasOwnProperty('set') || allowed_custom_attributes.hasOwnProperty('delete')">
```
fritterhoff: Think it makes sense to check for both. The existence and the length

fritterhoff (changes requested):

edumfa/static/components/user/views/user.details.html
cbergmann: sorry that I did not find the time to respond faster. You are right. We should check both here

PR #746, opened 2025-08-23
feat: add create timestamp to token

fritterhoff:
Add a new info to all tokens: Create timestamp

### DIFF
--- a/edumfa/lib/token.py
+++ b/edumfa/lib/token.py
@@ -1116,7 +1116,7 @@ def init_token(param, user=None, tokenrealms=None,
     use_timeshift = param.get("useTimeShift")
     if use_timeshift is not None and use_timeshift != "default":
         tokenobject.add_tokeninfo("useTimeShift", use_timeshift)
-
+    tokenobject.add_tokeninfo("created_at", datetime.datetime.now(tzlocal()).isoformat(sep=' ', timespec='microseconds'))
     # Safe the token object to make sure all changes are persisted in the db
     tokenobject.save()
     return tokenobject
--- a/edumfa/static/components/token/views/token.details.html
+++ b/edumfa/static/components/token/views/token.details.html
@@ -207,7 +207,7 @@
                         </span>
                         <span ng-hide="['certificate', 'pkcs12']
                         .indexOf(k)!=-1">
-                            {{ k }}: {{ v | truncate:30 }}
+                            {{ k }}: {{ v | truncate:35 }}
                         </span>
                     </li>
                 </ul>


edumfa/lib/token.py
```diff
@@ -1116,7 +1116,7 @@ def init_token(param, user=None, tokenrealms=None,
     use_timeshift = param.get("useTimeShift")
     if use_timeshift is not None and use_timeshift != "default":
         tokenobject.add_tokeninfo("useTimeShift", use_timeshift)
-
+    tokenobject.add_tokeninfo("created_at", datetime.datetime.now(tzlocal()).isoformat(sep=' ', timespec='microseconds'))
     # Safe the token object to make sure all changes are persisted in the db
```
aleyna72072, line 1120: Could you please also fix the grammar issue below? Safe -> Save

edumfa/lib/token.py
aleyna72072, line 1: grafik

It's a bit annoying that the timestamp gets cut off in the token detail view, would it be possible to change that?  
https://github.com/eduMFA/eduMFA/blob/main/edumfa/static/components/token/views/token.details.html#L210

edumfa/lib/token.py
```diff
@@ -1116,7 +1116,7 @@ def init_token(param, user=None, tokenrealms=None,
     use_timeshift = param.get("useTimeShift")
     if use_timeshift is not None and use_timeshift != "default":
         tokenobject.add_tokeninfo("useTimeShift", use_timeshift)
-
+    tokenobject.add_tokeninfo("created_at", datetime.datetime.now(tzlocal()).isoformat(sep=' ', timespec='microseconds'))
```
aleyna72072, line 1119: The function docstring says: "or update an existing token"  
Obviously `created_at` should not be updated when modifying a token. I in my testing it didn't change the timestamp, but as I don't know why that's the case this comment exists.

aleyna72072 (changes requested):
It'd be nice if it didn't cut off.

edumfa/lib/token.py
fritterhoff: Well. We call this function inside the init_token method. That should not be triggered more than once.

aleyna72072: https://github.com/eduMFA/eduMFA/blob/v2.8.0/edumfa/lib/token.py#L1033-L1044
Not really sure what is updated here, but it errors when testing. So it's probably fine.

edumfa/lib/token.py
aleyna72072: Maybe we don't need the full precision in the webui as an alternative to truncating less.

aleyna72072 (approved):

aleyna72072:
grafik

@fritterhoff Looks good, thanks!

PR #754, opened 2025-08-23
feat: replace crypt with passlib

fritterhoff:
The crypt package has been removed with Python 3.13. As a replacement, we used crypt_r, which is a drop-in replacement by just being a copy of the removed package. As a better permanent solution we migrate to passlib.

Since passlib has been unmaintained for a while, we need probably find a replacement for that too.

`edumfa-create-sqlidresolver-user` gets removed since we don't promote this tool in the docs/...


aleyna72072:
Background:  
https://peps.python.org/pep-0594/#crypt  
https://docs.python.org/3/library/crypt.html

edumfa/lib/resolvers/PasswdIdResolver.py
```diff
@@ -43,11 +43,10 @@
 from edumfa.lib.utils import to_bytes, convert_column_to_unicode
 from .UserIdResolver import UserIdResolver
 
-# Python 3.13 dropped crypt package, so we need to import crypt_r
-if sys.version_info >= (3, 13):
-    import crypt_r as crypt
-else:
-    import crypt
+# Python 3.13 dropped crypt package, so we need to migrate to passlib
```
Luc1412, line 46:
```suggestion
```
Think that can be removed

Luc1412:
```suggestion
```
Think that can be removed

edumfa/lib/resolvers/PasswdIdResolver.py
```diff
@@ -200,14 +199,36 @@ def checkPass(self, uid, password):
...
+
+            try:
+                if sha512_crypt.verify(password, cryptedpasswd):
+                    log.info("successfully authenticated user uid {0!s}".format(uid))
+                    return True
+            except ValueError as e:
+                pass
+
+            log.warning("user uid {0!s} failed to authenticate".format(uid))
+            return False
```
Luc1412, line 220:
```suggestion            
            handlers = [des_crypt, md5_crypt, sha256_crypt, sha512_crypt]
            for handler in handlers:
                try:
                    if handler.verify(password, cryptedpasswd):
                        log.info("successfully authenticated user uid {0!s}".format(uid))
                        return True
                except ValueError:
                    continue
            log.warning("user uid {0!s} failed to authenticate".format(uid))
            return False
```
We can shorten that

Luc1412 (changes requested):
lgtm, just some code improvements + we need to remove crypt_r also from the dependencies

fritterhoff:
Just noticed that there are two scripts also using crypt. Would consider dropping them completely?

aleyna72072:
> Just noticed that there are two scripts also using crypt. Would consider dropping them completely?

tools/edumfa-create-pwidresolver-user and tools/edumfa-create-sqlidresolver-user?  
At least the first one I use frequently and is also mentioned in the docs.

aleyna72072:
Instead of brute forcing, maybe:
```
from passlib.context import CryptContext
passlib_context = CryptContext(schemes=["bcrypt", "sha512_crypt", "sha256_crypt", "md5_crypt", "des_crypt"])
passlib_context.verify(password, cryptedpasswd)
```
The list above is not exhaustive but only an example, see https://passlib.readthedocs.io/en/stable/lib/passlib.hash.html#unix-hashes for the full list. I am unsure which schemes to support/list here. If we're going for compatibility, all of them (I did not find a way to not specify each one), if we're going for modern/secure then passlib has [recommendations](https://passlib.readthedocs.io/en/stable/narr/quickstart.html#the-options).

aleyna72072:
For `/tools/edumfa-create-pwidresolver-user`: https://gist.github.com/aleyna72072/896c8ce7ab10a5f74372398bc8c6ec7c

Not sure if the copyright header should get updated.

aleyna72072:
I think edumfa-create-sqlidresolver-user and its manpage could be deleted? I don't see it referenced anywhere.

aleyna72072 (changes requested):
See comments above.

tools/edumfa-create-pwidresolver-user
```diff
@@ -69,17 +65,14 @@ def main():
     if password == "":
         password = getpass.getpass("Please enter a password: ")
 
-    # without salt, the strongest available hash function will be selected
-    # and a random salt is created
-    encryptedPW = crypt.crypt(password)
+    passlib_context = CryptContext(schemes=["bcrypt"])
```
aleyna72072: Switch to "sha512_crypt", bcrypt hashing is broken with 5.0.0 and since passlib seems unmaintained it's unlikely to start working soon.
https://foss.heptapod.net/python-libs/passlib/-/issues/196
I tested with 4.3.0, my bad.

edumfa/lib/resolvers/PasswdIdResolver.py
```diff
@@ -213,14 +210,15 @@ def checkPass(self, uid, password):
...
                 raise NotImplementedError(err)
-            cp = crypt.crypt(password, cryptedpasswd)
-            log.debug("encrypted pass is {0!s}".format(cp))
-            if crypt.crypt(password, cryptedpasswd) == cryptedpasswd:
-                log.info("successfully authenticated user uid {0!s}".format(uid))
-                return True
-            else:
-                log.warning("user uid {0!s} failed to authenticate".format(uid))
-                return False
+            passlib_context = CryptContext(schemes=["bcrypt", "sha512_crypt", "sha256_crypt", "md5_crypt", "des_crypt"])
```
aleyna72072: Verifying a bcrypt password still works even on 5.0.0, but creates an exception in the log:
```
    version = _bcrypt.__about__.__version__
              ^^^^^^^^^^^^^^^^^
AttributeError: module 'bcrypt' has no attribute '__about__'

```
Maybe just remove it here too?

aleyna72072 (changes requested):
See below.
Passlib [seems unmaintained](https://foss.heptapod.net/python-libs/passlib/-/issues/187), so at some point we might need to switch this out, too. But since we use it in a lot of other places, this is probably fine for now.

fritterhoff:
Uff replacing passlib will cause some work...

aleyna72072:
[Ansible](https://github.com/ansible/ansible/issues/81949#issuecomment-1768772765) seems to think so, too :grimacing: .

aleyna72072:
@fritterhoff Can you please also fix the linting issues?

aleyna72072 (approved):
LGTM, but as Lucas said above: the dependency is still in the requirements.

Luc1412 (changes requested):
Looks good, but I think we should add reasoning to the description, why removing those 2 scripts

Luc1412 (approved):

PR #755, opened 2025-08-23
feat: better handling of broken aes yubikey

### LINKED ISSUE #732: Internal Server Error when using a faulty Yubikey in AES mode (wrong token length)
### Top-level intent

One of our clients used a Yubikey in AES-mode as second factor in eduMFA. When they tried to use this second factor at their shibboleth IdP, eduMFA responded with a Interal Server Error. It seems that the configuration of the HW-token was somehow faulty, because the error which was thrown at eduMFA said that the provided token had an unexpected length (see attached logfile). Therefore we assume that the actual issue with the token is located in the enrolled token or the token-configuration.
It would be nice if this error could be catched and a meaningfull errormessage would be delivered to the client. 

After re-enrolling the token the issue was gone but the old token-slot is still present and could be used to reproduce the issue and collect more information, if needed.

### Steps to reproduce

1. Login at shibboleth-IdP connected to eduMFA via fudiscr-Plugin using the faulty Yubikey

### Expected outcome

eduMFA should not responde with a 500 Internal Server Error but with a meaningfull errormessage. Something like "The provided token had an unexpected length. Please check your Token." 

### Actual outcome

eduMFA responded with an Internal Server Error and the login process was interrupted.

### Configuration

* **eduMFA version**: 2.8.0
* **Installation method**: from Ubuntu packages (edumfa-nginx)
* **Python version**: 3.12.3
* **Operating system**: Ubuntu 24.04
* **Webserver**: Nginx
* **Token database**: Mariadb

### Log file

[edumfa-token-length-expcetion.log](https://github.com/user-attachments/files/21835021/edumfa-token-length-expcetion.log)

fritterhoff:
as discussed in #732 handling of wrong aes authentication is not that good right now. instead of throwing an error return an errorcode and cancel the authentication process.


edumfa/lib/tokens/yubikeytoken.py
```diff
@@ -272,8 +273,11 @@ def check_otp(self, anOtpVal, counter=None, window=None, options=None):
         except KeyError:
             # The OTP value is no yubikey aes otp value and can not be decoded
             return -4
-
-        msg_bin = secret.aes_ecb_decrypt(otp_bin)
+        try:
+            msg_bin = secret.aes_ecb_decrypt(otp_bin)
+        except binascii.Error as e:
+            log.warning("handling token response for token {0} failed ({1})".format(serial, e))
```
aleyna72072, line 287: NIT: in other places the first letter of the log message is uppercase/the log message is a proper sentence

aleyna72072 (changes requested):
LGTM, but it'd be nice to have a unit test for this.

fritterhoff:
> LGTM, but it'd be nice to have a unit test for this.

Testing without any knowledge about the real error is pretty hard ;/

aleyna72072:
@fritterhoff I've just tested, the error comes from a secret being too short during enrollment. At least the new Yubico Authenticator does only accept strings exactly 32 characters long.  
Maybe instead (or additionally) enforce this exact length (for the secret, not the otp-length) on rollout?

aleyna72072:
Update: it is due to the secret being a wrong length. But since it also can create a ValueError, I hope to be able to formulate a patch which catches that (and enforces 32 characters in the frontend at least).

aleyna72072:
I've implemented an alternative to this at #820.

PR #756, opened 2025-08-23
feat: enable alembic logging

fritterhoff:
Until now alembic migrations did not output any log messages. The most of these are actually pretty helpful. 

-> Enable them in the default logging config and the provided config for docker. 

In case of the docker version we ship a default handler so the propagation has to be stopped to prevent duplicate log messages.

### DIFF
--- a/deploy/docker/logging.cfg
+++ b/deploy/docker/logging.cfg
@@ -18,6 +18,18 @@
             "handlers": ["stdhandler"],
             "level": "INFO"
         },
+        "alembic": {
+            "handlers": ["stdhandler"],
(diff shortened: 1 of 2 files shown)


Luc1412 (approved):

aleyna72072:
@fritterhoff Was it intentional to not enable this for the Ubuntu packages? https://github.com/eduMFA/eduMFA/blob/main/deploy/logging.cfg

fritterhoff:
Nope. Missed this special file.

aleyna72072:
Doesn't even seem to be used. I've created an issue.

PR #767, opened 2025-09-04
chore: Add ruff formatter and enforce formatting with gh action

Luc1412:
This PR introduced a project code style, that will also be enforced trough GitHub actions. This is the first step to leverage code quality and our first linting measure with more to come.

While black is probably the best known formatter, we decided for ruff. It's rules are 99,9%+ identical to black, while providing significant better performance. We also consider to implement further linting rules from ruff and even consider using uv by the same developer.

Besides changes by the formatter, following things have been changed:
- New linting action, that checks if the code has been formatted with ruff
- New linting dependency extra for installed these dependencies 

> [!IMPORTANT]


fritterhoff:
As discussed. We should have a short look onto the translations.

fritterhoff (approved):

PR #776, opened 2025-09-08
update mysql charset for new deployments

### LINKED ISSUE #728: MySQL character set
### Description
Currently in our Ubuntu packages we use `utf8` as the mysql character set. This is [currently an alias](https://dev.mysql.com/doc/refman/9.4/en/charset-unicode-utf8.html) for `utf8mb3`, which is [deprecated](https://dev.mysql.com/doc/refman/9.4/en/charset-unicode-sets.html) and to be deleted in a future release, where `utf8` [will](https://dev.mysql.com/doc/refman/9.4/en/charset-unicode-utf8.html) mean `utf8mb4` instead. They recommend setting `utf8mb4` explicitly still to avoid confusion.  

### Open Questions
- How do you migrate from `utf8mb3` to `utf8mb4`? Can we make it automatic in a migration?
  + "changing character sets can be a complex and time-consuming task", [MySQL docs](https://dev.mysql.com/doc/refman/9.4/en/charset-unicode-utf8mb3.html)
- Do we even need to do it? Is there a problem for users when `utf8` suddenly means utf8mb4 instead?
  + If not, we could change this  for new installations already at least.
  + Did this already happen for deployments which don't specify a charset when flask-sqlalchemy [changed it](https://github.com/pallets-eco/flask-sqlalchemy/pull/1014)?

### Places
Relevant places found by grepping (are there any more?):
- https://github.com/eduMFA/eduMFA/blob/main/deploy/ubuntu-server/edumfa-apache2.postinst#L165
- https://github.com/eduMFA/eduMFA/blob/main/deploy/ubuntu-server/edumfa-nginx.postinst#L166
- https://github.com/eduMFA/eduMFA/blob/main/migrations/env.py#L23-L35
  + I don't understand what this does.
  + Flask-SQLAlchemy [already](https://github.com/pallets-eco/flask-sqlalchemy/pull/1014) switched over.

aleyna72072:
Fixes #728 at least for new deployments. I haven't found a solution for existing deployments.

### DIFF
--- a/deploy/ubuntu-server/edumfa-apache2.postinst
+++ b/deploy/ubuntu-server/edumfa-apache2.postinst
@@ -162,7 +162,7 @@ create_database() {
     mysql -u "$USER" --password="$PASSWORD" -e "create database edumfa;" || true
     mysql -u "$USER" --password="$PASSWORD" -e "create user 'edumfa'@'localhost' IDENTIFIED BY '$NPW';"
     mysql -u "$USER" --password="$PASSWORD" -e "grant all privileges on edumfa.* to 'edumfa'@'localhost';"
-    echo "SQLALCHEMY_DATABASE_URI = 'mysql+pymysql://edumfa:$NPW@localhost/edumfa?charset=utf8'" >> /etc/edumfa/edumfa.cfg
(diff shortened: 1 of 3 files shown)


aleyna72072:
This should also fix the migrations always using utf8 even if specified differently in the config.

Luc1412 (approved):

PR #781, opened 2025-09-12
image: enable simple configuration via env vars

aleyna72072:
Ship a default edumfa.cfg which can be configured with environment variables.


doc/installation/docker.rst
```diff
@@ -21,73 +21,94 @@ Docker Compose
...
+         EDUMFA_ADMIN_PASS: ${EDUMFA_ADMIN_PASS}
+         EDUMFA_UI_DEACTIVATED: ${EDUMFA_UI_DEACTIVATED}
        depends_on:
          mariadb:
            condition: service_healthy
 
+   networks:
+     edumfa-net:
+       name: edumfa-network
+       driver: bridge
```
Luc1412: Do we need an extra network? Usually the default network should work fine.

doc/installation/docker.rst
```diff
@@ -21,73 +21,94 @@ Docker Compose
...
+   EDUMFA_SECRET_KEY="Test"
+   EDUMFA_PEPPER="Never know..."
+
+   EDUMFA_ADMIN_USER="admin"
+   EDUMFA_ADMIN_PASS="Passwort123"
+   SUPERUSER_REALM="super,administrators"
+   EDUMFA_UI_DEACTIVATED="False"
+
+
+For production you should replace the passwords and secrets with your own values.
```
Luc1412: I'd might suggest a different documentation approach. Instead of providing example values something like this layout might be better. It also should include whats required and whats optional:
- `MARIADB_USER`: The MariaDB User
- `MARIADB_PASSWORD`: The MariaDB password
...

doc/installation/docker.rst
```diff
@@ -21,73 +21,94 @@ Docker Compose
...
+For production you should replace the passwords and secrets with your own values. 
 
 To start eduMFA using Docker Compose, run:
 
 .. code-block:: bash
 
    docker compose up -d
 
-For more information on using eduMFA, please refer to :ref:`first_steps`.
+Alternatively, you can mount your own `edumfa.cfg` instead of configuring eduMFA via environment variables. For more information on using eduMFA, please refer to :ref:`first_steps`.
```
Luc1412: This comment regarding alt. config should be after .env explaination

deploy/docker/edumfa.py
```diff
@@ -0,0 +1,34 @@
+from os import getenv
+
+
+def _getenv(key, default="") -> str:
```
Luc1412:
```suggestion
def _getenv(key: str, default: str | None = None) -> str:
```
Added type hints and `None` as default seems more appropriate.

deploy/docker/edumfa.py
```diff
@@ -0,0 +1,34 @@
...
+    :return: The environment variable's value or the default if given.
+    :raises Exception: If the variable is not set and there is no default
+        value.
+    """
+    value = getenv(key)
+    value = value if value else default
+    if not value:
+        raise Exception(f"Environment variable '{key}' not set! Can't start...")
+    else:
+        return value
```
Luc1412:
```suggestion
    :raises ValueError: If the variable is not set and there is no default
        value.
    """
    value = getenv(key, default)
    if not value:
        raise Exception(f"Environment variable '{key}' not set! Can't start...")
    return value
```

deploy/docker/edumfa.py
```diff
@@ -0,0 +1,34 @@
...
+SQLALCHEMY_DATABASE_URI = f"{_getenv('DB_DRIVER')}://{_getenv('DB_USER')}:{_getenv('DB_PASSWORD')}@{_getenv('DB_HOSTNAME')}/{_getenv('DB_DATABASE')}"
+SUPERUSER_REALM = _getenv("SUPERUSER_REALM", "super,administrators").split(",")
+SECRET_KEY = _getenv("SECRET_KEY")
+EDUMFA_PEPPER = _getenv("EDUMFA_PEPPER")
+EDUMFA_ENCFILE = "/etc/edumfa/enckey"
+EDUMFA_AUDIT_KEY_PRIVATE = "/etc/edumfa/private.pem"
+EDUMFA_AUDIT_KEY_PUBLIC = "/etc/edumfa/public.pem"
+EDUMFA_LOGFILE = "/var/log/edumfa/edumfa.log"
+EDUMFA_LOGCONFIG = "/etc/edumfa/logging.yml"
+EDUMFA_UI_DEACTIVATED = _getenv("EDUMFA_UI_DEACTIVATED", "False") == "True"
```
Luc1412, line 32:
```suggestion
EDUMFA_UI_DEACTIVATED = bool(_getenv("EDUMFA_UI_DEACTIVATED", "False"))
```

Luc1412 (changes requested):
PR looks good, but got some small suggestions.

doc/installation/docker.rst
aleyna72072: I don't think so, I kept it like it was before. I don't have a horse in this race, should I remove it?

deploy/docker/edumfa.py
aleyna72072:
```
>>> bool("False")
True
>>> bool("false")
True

```
Am I doing something wrong or is this not working as intended?

doc/installation/docker.rst
aleyna72072: done

doc/installation/docker.rst
aleyna72072: done

deploy/docker/edumfa.py
aleyna72072: done

deploy/docker/edumfa.py
aleyna72072: done

doc/installation/docker.rst
aleyna72072: Removed it.

deploy/docker/edumfa.py
Luc1412: mb, that doesn't work since it evaluates the strings lenght > 0

doc/installation/docker.rst
Luc1412: I think it would be better to remove both references in docs and example compose, to keep it as simple as possible

aleyna72072: Missed that; done!

Luc1412 (approved):

PR #789, opened 2025-09-22
Implement Isort linting ruleset

Luc1412:
This PR implements the first linting ruleset `I` (Isort).

Addionally it updates ruff related stuff including:
- ruff-action for linting CI
- update ruff

> [!IMPORTANT]


fritterhoff (approved):
Looks good to me (even though I am not such a big fan of the grep inside the github actions but have no better idea)

PR #796, opened 2025-09-23
Add support for Python 3.14 and drop support for Python 3.9

Luc1412:
This PR adds support for Python 3.14 and drops Support for Python 3.9.

This includes:
- Workflows now run on Python 3.14
- Tests now include Python 3.14, but Python 3.9 was removed
- Ubuntu Image now defaults to >= 3.10
- Docker image now uses Python 3.14
  - Addionally Debian Trixie instead of Bookworm is used
- Docs and project metadata has been updated
- Test `test_36_imagefile_to_dataimage` has been fixed
  - Python 3.14 now detects Yaml files -> the test doesn't fail anymore as exspected
  - Now using a `.cfg` file from testdata


tests/test_lib_utils.py
```diff
@@ -1175,6 +1175,6 @@ def test_36_imagefile_to_dataimage(self):
         self.assertEqual("", dataimage)
 
         # Try to read a grazy file with an unknown mime-type
-        file = "./tests/testdata/logging.yml"
+        file = "./tests/testdata/logging.cfg"
```
aleyna72072, line 1178: Curious: Why this change?

Luc1412: I mentioned this in the PR overview. It's because of Py3.14 now supporting more Mime Types.

aleyna72072 (approved):
LGTM

fritterhoff (approved):

PR #803, opened 2025-09-26
chore: implement Pyupgrade linting ruleset

Luc1412:
This PR implements the first linting ruleset `UP` (Pyupgrade). Besides using Pyupgrade's rules, manually migration to f-strings was done.

> [!IMPORTANT]


Luc1412:
With this PR I found an issue with `test_02_get_auth_item` in `test_lib_applications.py`. User in the log message is resolved by `options` and not by `filter_param`.

How should we correct the test

aleyna72072:
> With this PR I found an issue with `test_02_get_auth_item` in `test_lib_applications.py`. User in the log message is resolved by `options` and not by `filter_param`.

Can you elaborate? Is the test failing? If so, why is it not failing in the CI?  

I just took a look at it, and I _do wonder_ why the second Idefix in the parenthesis is expected to be there in the test. Shouldn't options be an empty dict there? Am I missing something?  

https://github.com/eduMFA/eduMFA/blob/v2.9.0/tests/test_lib_applications.py#L72-L74
https://github.com/eduMFA/eduMFA/blob/v2.9.0/tests/test_lib_applications.py#L76-L79
https://github.com/eduMFA/eduMFA/blob/v2.9.0/edumfa/lib/applications/ssh.py#L58-L66
https://github.com/eduMFA/eduMFA/blob/v2.9.0/edumfa/lib/applications/ssh.py#L97-L102

edumfa/api/auth.py
```diff
@@ -244,7 +239,7 @@ def get_auth_token():
     # the realm parameter has precedence! Check if it exists
     if realm_param and not realm_is_defined(realm_param):
         raise AuthError(
-            _("Authentication failure. Unknown realm: {0!s}.".format(realm_param)),
+            _("Authentication failure. Unknown realm: {0!s}.").format(realm_param),
```
aleyna72072, line 242:
```suggestion
            _(f"Authentication failure. Unknown realm: {realm_param!s}."),
```
Did not check if it's already a string though.

edumfa/lib/tokens/totptoken.py
```diff
@@ -522,7 +507,7 @@ def _autosync(self, hmac2Otp, anOtpVal):
                 self.set_tokeninfo(info)
 
             else:
-                log.debug("setting otp1c: {0!s}".format(res))
+                log.debug(f"setting otp1c: {res}")
```
aleyna72072, line 510: There are quite some changes like this to f-strings (yay), where the conversions are dropped. Did you manually check if the data types still fit or is there some kind of automated assurance?  

(Reading https://peps.python.org/pep-0498/#s-r-and-a-are-redundant I get the impression they are still necessary.)

Luc1412: Everything in fstrings gets converted to a string. So !s is unnecessary.

aleyna72072: So !s is not needed, but !a and !r? I tried to find documentation on this but didn't struck gold.

edumfa/api/auth.py
Luc1412: That's generally a mistake. Translated strings still need to use .format() but outside the translate function e.g. _().format()

aleyna72072: Ah, good catch!

edumfa/lib/tokens/totptoken.py
Luc1412: Yeah, since f-strings always apply str() or !s to all vars. However !a is equivalent of ascii() and !r repr()

Luc1412:
> Can you elaborate? Is the test failing? If so, why is it not failing in the CI?

With this PR I identified a formatting issue in https://github.com/eduMFA/eduMFA/pull/803/changes#diff-0623d0baef87aa9f29d96137d2849d4b298e19de15e39f827c96ede2d7822347L98-L101

They mistakenly always inserted the first var ({0}) twice instead of also using {1}. My changes resolves it and restore intended behaviour. However this breaks the test and changes the ouput.

edumfa/lib/applications/ssh.py
aleyna72072, line 62:
```suggestion
        :param serial: the serial number of the token.
        :param challenge: A challenge for the authitem.
        :param options: The MachineTokenOptions of the token.
        :param filter_param: A dictionary whose value for the key "user" has to
                             match the user the token is assigned to in the
                             MachineTokenOptions.
```

aleyna72072: I am not sure what "challenge" does, I stole that description from `lib/machine.py`. `api/machine.py` says:
>    :param challenge: A challenge for which the authentication item is
>        calculated. In case of the Yubikey this can be a challenge that produces
>        a response. The authentication item is the combination of the challenge

But without further digging I'm not sure what it means. But the parameter description is probably not wrong at least.

tests/test_lib_applications.py
```diff
@@ -75,7 +76,7 @@ def test_02_get_auth_item(self):
             self.assertFalse(auth_item)
             mock_log.assert_called_with(
                 "The requested user Idefix does "
-                "not match the user option (Idefix) of the SSH application."
+                "not match the user option (None) of the SSH application."
```
aleyna72072:
```suggestion
                "not match the user option (cornelius) of the SSH application."
```
see below

tests/test_lib_applications.py
aleyna72072:
```suggestion
                "sshkey",
                serial,
                filter_param={"user": "Idefix"},
                options={"user": "cornelius"},
```
I think this is how the original unit test was meant.

tests/test_lib_applications.py
```diff
@@ -75,7 +76,7 @@ def test_02_get_auth_item(self):
...
             )
```
aleyna72072, line 83:
```suggestion
            )
        # Get with missing MachineTokenOptions, meaning the attached user can't
        # be checked.
        with mock.patch("logging.Logger.debug") as mock_log:
            auth_item = SSHApplication.get_authentication_item(
                "sshkey",
                serial,
                filter_param={"user": "Idefix"},
            )   
            self.assertFalse(auth_item)
            mock_log.assert_called_with(
                "The requested user Idefix does "
                "not match the user option () of the SSH application."
            )
```
Suggestion: also make sure that not passing the `options` parameter produces the correct result.

aleyna72072 (approved):
This is something I look forward to everytime I see the encoding line when I open a file! I just skimmed through the files, LGTM.
Since it is touching a lot of code, maybe also get a second review? Maybe @fritterhoff ?

aleyna72072:
> Besides using Pyupgrade's rules, manually migration to f-strings was done.

Also kudos for that, that must have taken forever? Glad to see that done.

fritterhoff (approved):

PR #804, opened 2025-09-26
fix migration 414079706620 for mysql upgrades

aleyna72072:
When applying the 414079706620 migration as an upgrade on MySQL, both `id` columns lose their AUTO_INCREMENT. 
This PR fixes that as described in this issue: https://github.com/sqlalchemy/alembic/issues/413

Strangely this does not happen on `edumfa-manage create_tables` (Maybe because the type is already bigint? Not sure when/why it is bigint or int before this migrations).  
On MariaDB the database gets created with sequences, meaning there is no AUTO_INCREMENT to be lost.

### DIFF
--- a/migrations/versions/414079706620_.py
+++ b/migrations/versions/414079706620_.py
@@ -31,6 +31,7 @@ def upgrade():
         existing_type=sa.Integer(),
         type_=BigIntegerType,
         existing_nullable=False,
+        existing_autoincrement=True,
     )
     print("Setting challenge.id to bigint")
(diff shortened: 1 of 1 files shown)


fritterhoff (approved):

aleyna72072:
This should be "more correct", but seems to behave the same (tested on mysql and psql). If you could approve it again, it would be great to get this in before the next release!
Also feel free to merge it as I am out of office next week.

fritterhoff (approved):

PR #816, opened 2025-10-06
chore: lint & format scripts with no .py ending

Luc1412:
This PR adds scripts with no .py ending to ruff linting and formatting.


aleyna72072 (approved):
LGTM

PR #822, opened 2025-10-10
feat: rootless docker image

### LINKED ISSUE #1170: feat(image): run as unprivileged user
**Description.**

I want to run the image without root privileges since our Kubernetes Platform has very strict security requirements. We need to set the following `securityContext` on a container:

```
    securityContext:
      allowPrivilegeEscalation: false
      capabilities:
        drop:
        - ALL
      runAsGroup: 1000
      runAsNonRoot: true
      runAsUser: 1000
      seccompProfile:
        type: RuntimeDefault
```

While this works with the current container image, it would be nice to have a non-root user be the default. 
This would also remove the shell promt with `I have no name!@edumfa-59b749bd65-zqdf8:/opt/edumfa$` when opening a shell in the container. This is purely cosmetic, though.

aleyna72072:
Makes the image rootless. As discussed, this is a breaking change. I've added instructions which should help users to migrate.  
A community member also created a similar PR: #1171

resolves #1170

### DIFF
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -5,6 +5,24 @@ GEN_PWD="$(openssl rand -base64 42)"
 EDUMFA_ADMIN_USER="${EDUMFA_ADMIN_USER:-admin}"
 EDUMFA_ADMIN_PASS="${EDUMFA_ADMIN_PASS:-$GEN_PWD}"
 
+# Check that necessary files are readable by the current user.
+function check_config_value_paths_for_read() {
+  key="$1"
+  value="$(edumfa-manage -q config get_value "$key" 2>/dev/null)"
+  if [ ! -r "$value" ]; then
+    echo "Can't access the file at '$value'. Please make sure it exists and is readable by uid 2000."
+    echo "If you are upgrading from an older version of eduMFA, it might be necessary to change the file owner:"
+    echo "1. Make sure eduMFA is stopped."
+    echo "2. Change ownership in all volumes. Example for a volume mounted at /etc/edumfa/:"
+    echo "\$ docker run -u root -v \$MY_VOLUME:/etc/edumfa/ ghcr.io/edumfa/edumfa:latest chown -R 2000:2000 /etc/edumfa/"
+    echo "If you want to avoid downtime you could use a cloned volume instead."
+    exit 1
+  fi
+}
+check_config_value_paths_for_read EDUMFA_ENCFILE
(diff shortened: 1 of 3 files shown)


aleyna72072:
Maybe the new command `edumfa-manage config get_value` can be also used to remove the need for the manual parsing in https://github.com/eduMFA/eduMFA/commit/9c864861a694124c860bce6c0ea0def63b469287 ?  

EDIT: use https://github.com/eduMFA/eduMFA/pull/1155/changes/929032b12f997fb70bd7f5bf81ac412eeebd4344 instead, it has tests

deploy/docker/entrypoint.sh
```diff
@@ -1,135 +1,76 @@
...
+# Check that necessary files are readable by the current user.
+function check_config_value_paths_for_read() {
+  key="$1"
+  value="$(edumfa-manage -q config get_value "$key" 2>/dev/null)"
+  if [ ! -r "$value" ]; then
+    echo "Can't access the file at '$value'. Please make sure it exists and is readable by uid 2000."
+    echo "If you are upgrading from an older version of eduMFA, it might be necessary to change the file owner:"
+    echo "1. Make sure eduMFA is stopped."
+    echo "2. Change ownership in all volumes. Example for a volume mounted at /etc/edumfa/:"
+    echo "\$ docker run -u root -v \$MY_VOLUME:/etc/edumfa/ ghcr.io/edumfa/edumfa:latest chown -R 2000:2000 /etc/edumfa/"
```
fritterhoff, line 17: Prop. that would not work. As soon as we migrate to a root less image the edumfa latest image would not have the permissions to fix that. So either fixing directly on the shell would be required or using any other image + mounting the volume :)

aleyna72072: In my testing it works fine:
```
$ cat Dockerfile 
FROM ghcr.io/edumfa/edumfa:2.9.5
USER nobody
$ docker build . -t x:1
[...]
$ docker run --rm -it x:1 bash
nobody@d5d8c86740ae:/opt/edumfa$ whoami
nobody
nobody@d5d8c86740ae:/opt/edumfa$ exit
exit
$ docker run -u root --rm -it x:1 bash
root@118ec5e1e867:/opt/edumfa# whoami
root
 ```
 
 But I could have absolutely missed something?

aleyna72072: But sure, we could just use a Busybox image. If all planets are in order and we actually somehow get distroless into 3.0.0 then this would be mandatory-ish anyway.

PR #836, opened 2025-10-16
add policy to force a timeshift setting

### LINKED ISSUE #655: force timeshift policy
When rolling out a TOTP token, you can hide technical details by setting user policies (e.g. totp_force_server_generate). Currently, there is no policy to hide totp_timeshift. 

Image

A policy setting akin to totp_force_server_generate to avoid confusing users if not necessary. That setting would force using the default in the TOTP token setting, without exposing that option to the user.

**Describe alternatives you've considered**
Custom CSS, [documentation](https://edumfa.readthedocs.io/en/v2.8.0/faq/customization.html):
```
div[ng-hide="checkRight('totp_timeshift')"] {
    display: none;
}
```

**Implementation**
If I'll get the time granted at some point, I'll take a shot at it.

aleyna72072:
Users often want to hide the timeshift input to not confuse users. This patch adds a policy to enforce default/True/False, which hides the input.
resolves #655


Luc1412 (changes requested):
Looks good, but translations need to be updated

aleyna72072:
> Looks good, but translations need to be updated

Done, the commits are huge though because of the refactoring commits I assume. Would it be possible to add a check for PRs to ensure translations for the server and webui are up to date?

aleyna72072:
@Luc1412 updated translations

Luc1412 (approved):

PR #845, opened 2025-10-24
fix: handle token reset for passkey auth

fritterhoff:
A sucessfull authentication does not reset the other failcounters for the tokens. 

This PR fixes the behaviour.

### DIFF
--- a/edumfa/lib/tokens/webauthntoken.py
+++ b/edumfa/lib/tokens/webauthntoken.py
@@ -573,7 +573,7 @@ class WEBAUTHNGROUP:
 
     WEBAUTHN = "WebAuthn"
 
-def reset_all_user_tokens_passkey(user):
+def reset_all_user_tokens_passkey(user) -> None:
     # Gather all tokens of the user that are not registration tokens and reset the failure counter
     reset_all = Match.user(
         g,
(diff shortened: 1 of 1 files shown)


edumfa/lib/tokens/webauthntoken.py
```diff
@@ -573,6 +573,28 @@ class WEBAUTHNGROUP:
...
+        ]
+        log.debug("Reset failcounter of all tokens of {0!s}".format(user))
+        for tok_obj_reset in tokens:
+            try:
+                tok_obj_reset.reset()
+            except Exception:
+                log.warning(
+                    "Could not reset failure for token {0!s} of user {1!s}".format(tok_obj_reset.serial, user)
+                )
+
```
Luc1412:
```suggestion
def reset_all_user_tokens_passkey(user) -> None:
    # Gather all tokens of the user that are not registration tokens and reset the failure counter
    reset_all = Match.user(
        g,
        scope=SCOPE.AUTH,
        action=ACTION.RESETALLTOKENS,
        user_object=user if user else None,
    ).policies()
    if not reset_all:
        return
    tokens = get_tokens_from_serial_or_user(None, user=user)
    log.debug("Reset failcounter of all tokens of {0!s}".format(user))
    for tok_obj_reset in tokens:
        if tok.get_class_type() in ["registration"]:
            continue
        try:
            tok_obj_reset.reset()
        except Exception:
            log.warning(
                "Could not reset failure for token {0!s} of user {1!s}".format(tok_obj_reset.serial, user)
            )

```
Removes unnecessary nesting

Luc1412 (changes requested):
Looks good so far. Just some code improvements + it would be nice if the PR description could include more information about what exactly is getting fixed here

edumfa/lib/tokens/webauthntoken.py
```diff
@@ -573,6 +573,28 @@ class WEBAUTHNGROUP:
 
     WEBAUTHN = "WebAuthn"
 
+def reset_all_user_tokens_passkey(user) -> None:
```
aleyna72072, line 576: No need to, but could you maybe add a note here like:
`# XXX: this should use the normal decorator instead`

So in the future it might get fixed properly.

Luc1412 (approved):

PR #853, opened 2025-10-28
ci: rework github workflows

Luc1412:
This pr rework and simplify the github workflows. 
### General Changes
- Replaced common python tools for managing dependencies, building dist and publishing the package with the all-in-one tool `uv`
  - This results in simplified workflows and faster speeds 
- Instead of using 3.14, the min. supported Python version is used for completing workflows
  - This is currently Python 3.10, except for docs where we use Python 3.11
- Names of workflows, jobs and steps have been unified
### Workflow changes
- `bandit` check has been removed. If we want to use Bandit, we can use the ruff bandit ruleset (`S`)
  - Currently there would be many issues that need to be resolved, so it's better part of another pr 
- `codeql` workflow has been merged into `lint` workflow
- `codespell` workflow has been merged into `lint` workflow
- `docker` workflow has been optimized, by adding specific paths to only run the action
- `docs` have been moved into `build` workflow
- `greetings` have been removed. Issue/PR templates solve this matter already
- `release` workflow was completely overhauled with uv tooling
- `python` workflow has been more appropriately renamed to `test` and was completely overhauled
  - Instead of using a pre check with filter, the native github actions filter is used
  - Instead of testing all versions 3.10-3.14 only min and max will be tested
- `ubuntu` workflow has been optimized, by adding specific paths to only run the action
- `translations` workflow has been added to check frontend and backend translations

> [!IMPORTANT]
> We need to updated our trusted publishing, since the workflow file name changed


aleyna72072:
Things from today's meeting which would be nice to have:

- shellsheck for scripts (e.g. entrypoint.sh)
- some linter for the html templates (xmllint?)
- check if translations are up to date

The only thing missing I can think of right now would be to also lint rst/the documentation. I don't know if there's a good tool or if it makes sense.

fritterhoff:
@Luc1412 as discussed:
https://babel.pocoo.org/en/latest/cmdline.html#extract

```
  --no-location         do not include location comments with filename and
                        line number
  --add-location=ADD_LOCATION
                        location lines format. If it is not given or "full",
                        it generates the lines with both file name and line
                        number. If it is "file", the line number part is
                        omitted. If it is "never", it completely suppresses
                        the lines (same as --no-location).
```

.github/workflows/docker.yml
```diff
@@ -1,40 +1,44 @@
...
+      - 'edumfa/**/*.py'
+      - 'tools/**'
+      - '.docerignore'
+      - 'Dockerfile'
+      - 'edumfa-manage'
+      - 'pyproject.toml'
+      - 'requirements.txt'
+      - 'setup.py'
   pull_request:
-  workflow_dispatch:
```
fritterhoff, line 6: Manual building of docker images was pretty usefull for testing

.github/workflows/docker.yml
```diff
@@ -48,13 +52,12 @@ jobs:
...
             type=semver,pattern={{major}}.{{minor}}
             type=semver,pattern={{major}},enable=${{ !startsWith(github.ref, 'refs/tags/v0.') }}
-            type=raw,value=latest,enable=${{ startsWith(github.ref, 'refs/tags/') }}
-
+            type=raw,value=latest,enable=${{ github.ref_type == 'tag' }}
       - name: Build and push
         uses: docker/build-push-action@v6
         with:
           context: .
-          push: ${{ github.event_name != 'pull_request' && (startsWith(github.ref, 'refs/tags/') || github.ref_name == github.event.repository.default_branch) || github.event_name == 'workflow_dispatch' }}
```
fritterhoff, line 57: In case we delete the dispatch option we can remove that here also.

fritterhoff (changes requested):

.github/workflows/docker.yml
Luc1412: I believe building the image locally might be the better choice here. Is there anything against doing testing fully locally?

Luc1412:
Regarding comments here:

@aleyna72072 I added the translations workflow. I think we gonna handle further linting within separated PRs, since it probably include big diffs.

@fritterhoff I think we update the translations within another PR.

fritterhoff:
OK. Then I will build the images locally. Was only a bit lazy 🙃

fritterhoff (approved):

PR #854, opened 2025-10-28
image: support template customization

aleyna72072:
The currently recommended way by the documentation to customize templates requires changing the reverse proxy potentially outside of the eduMFA compose file.  
This patch documents a way for users to instead just mount them over existing files. ~It requires us to update the symlink if we update the Python version.~

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -36,6 +36,10 @@ COPY ./deploy/gunicorn/edumfaapp.py /opt/edumfa/app.py
 # Create directory for user scripts
 RUN mkdir -p /opt/edumfa/user-scripts
 
+# Link the edumfa package at a predicatable position for template overriding.
+RUN python_package_path="$(python3 -c 'import site; x=site.getsitepackages(); assert len(x) == 1; print(x[0])')" && \
+    ln -s "${python_package_path}/edumfa/" "/opt/edumfa/edumfa-package"
+
 EXPOSE 8000
 HEALTHCHECK --interval=5s --timeout=3s --start-period=60s --retries=2 CMD curl --fail http://localhost:8000/ || exit 1
 WORKDIR /opt/edumfa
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,4 +1,6 @@
 # Build stage
+# Remember to update the symbolic link at the bottom when updating the Python
+# version!
```
Luc1412:
```suggestion
```

Dockerfile
```diff
@@ -36,6 +38,9 @@ COPY ./deploy/gunicorn/edumfaapp.py /opt/edumfa/app.py
 # Create directory for user scripts
 RUN mkdir -p /opt/edumfa/user-scripts
 
+# Update this path when updating the Python version!
+RUN ln -s /usr/local/lib/python3.14/site-packages/edumfa/ /opt/edumfa/edumfa-package
```
Luc1412:
```suggestion
RUN ln -s $(python3 -c "import site; print(site.getsitepackages()[0])")/edumfa/ /opt/edumfa/edumfa-package
```
Think that should solve the version dependence.

Luc1412 (changes requested):
Looks good, but I think we can bypass the version dependence in the docker image layers

Dockerfile
aleyna72072: That's a good idea! I've also added an assertion to avoid us silently breaking it. Not the prettiest though..

Dockerfile
aleyna72072: done

Dockerfile
fritterhoff: Do we really want to link the complete directory or only the static files?

aleyna72072: Is there an advantage to only linking the static dir?
If we link the complete directory users could use this for easy monkey-patching. Not saying we'd support or mention it, but it might come in handy for someone :smile: .

fritterhoff (approved):

Dockerfile
aleyna72072: agreed during meeting to leave as-is

aleyna72072:
@Luc1412 Not sure how I can merge it when your approval is still missing. Could you add your approval please? :slightly_smiling_face:

Luc1412 (approved):

PR #866, opened 2025-10-31
fix: stamp on create and include missing command

fritterhoff:
As discussed: use stamp in command and fix the missing import


aleyna72072 (approved):
lgtm

PR #869, opened 2025-10-31
refactor: restructure frontend translations

Luc1412:
This PR updates several parts of the frontend:
- translation related files have been moved in `edumfa/static`
  - translation files where moved from `po` to `edumfa/static/locales`
  - `Gruntfile.js` has been moved in the static dir
  - Merge root `package.json` with the one located in `edumfa/static`
- Migrate from `npm` to `bun` as package manager
- Update static files and local message


fritterhoff (commented):
I thought it was possible to manually tell git that the files where only moved....

Luc1412:
> I thought it was possible to manually tell git that the files where only moved....

Unfortunately I couldn't find a solution for that. 3bca80503b0c76f7fdffe85e4245854a652a8cd2 still displays the files as moved, while the commit afterwards cause the diff to break. I just retried with `git mv` but there is no chance of keeping the diff small.

fritterhoff (approved):

PR #880, opened 2025-11-12
image: add ability to read values from files

aleyna72072:
Storing secrets as environment variables can be dangerous (e.g. by accidentally dumping the environment or introducing other processes into the container). This pull request enables to pass each option as a path instead by appending `_FILE` to the variable name. This should ease integration in Kubernetes or secret management tools like [systemd's ](https://systemd.io/CREDENTIALS/).


deploy/docker-example/docker-compose.yml
```diff
@@ -26,11 +26,12 @@ services:
       - ./edumfa-setup.sh:/opt/edumfa/user-scripts/01-setup.sh:ro
       - ./example-policy.py:/opt/edumfa/policy.py:ro
       - ./my-login.html:/opt/edumfa/edumfa-package/static/components/login/views/login.html
+      - ./db_password.txt:/etc/edumfa/db_password.txt:ro
     environment:
       DB_DRIVER: ${DB_DRIVER}
       DB_HOSTNAME: mariadb
       DB_USER: ${MARIADB_USER}
-      DB_PASSWORD: ${MARIADB_PASSWORD}
+      DB_PASSWORD_PATH: /etc/edumfa/db_password.txt
```
Luc1412:
```suggestion
      DB_PASSWORD_FILE: /etc/edumfa/db_password.txt
```

deploy/docker/edumfa.py
```diff
@@ -20,14 +34,34 @@ def _getenv(key: str, default: str | None = None) -> str:
...
+        exist.
+    :return: The variable's value or the default if given.
+    :raises ValueError: If the variable is not set and there is no default
+        value.
+    """
+    variable_with_path_value = getenv(f"{key}_PATH")
+    if variable_with_path_value:
+        return get_content_from_file(variable_with_path_value)
+    else:
+        return _getenv(key, default)
```
Luc1412:
```suggestion
def get_var(key: str, default: str | None = None) -> str:
    """
    Decide whether to get the variable from the environment or a file,
    depending on if there is a environment variable with a _FILE-suffix
    which overrides the environment variable without it.

    :param key: The environment variable to get.
    :param default: The value to return if the environment variable doesn't
        exist.
    :return: The variable's value or the default if given.
    :raises ValueError: If the variable is not set and there is no default
        value.
    """
    path = getenv(f"{key}_FILE")
    if path:
        return get_content_from_file(path)
    return _getenv(key, default)
```
The naming convention of secret files is `_FILE` not `_PATH`

doc/installation/docker.rst
```diff
@@ -83,7 +83,13 @@ The `.env` file should contain the following variables:
 - SUPERUSER_REALM: which realms should be superuser realms (optional)
 - EDUMFA_UI_DEACTIVATED: whether to disable the WebUI (optional)
 
-For production you should replace the passwords and secrets with your own values. Alternatively, you can mount your own `edumfa.cfg` instead of configuring eduMFA via environment variables.
+You can also add a "_PATH" suffix to each variable name and pass a path to read the value from a file instead. For example instead of passing `SECRET_KEY`:
+
+.. code-block:: bash
+
+   SECRET_KEY_PATH: /etc/edumfa/secret_key.txt
```
Luc1412:
```suggestion
You can also add a "_FILE" suffix to each variable name and pass a path to read the value from a file instead. For example instead of passing `SECRET_KEY`:

.. code-block:: bash

   SECRET_KEY_FILE: /etc/edumfa/secret_key.txt
```

deploy/docker/edumfa.py
```diff
@@ -1,7 +1,21 @@
...
+
+    :param path: The path to read the file from.
+    :return: The content of the file stripped of newlines.
+    :raises FileNotFoundError: If the file at path can't be found.
+    """
+    with open(path) as fp:
+        content = fp.read()
+    # remove any leading/trailing newlines
+    content = content.strip("\n")
+    return content
```
Luc1412:
```suggestion
    """Return the normalized content from a path.

    :param path: The path to read the file from.
    :return: The content of the file stripped of newlines.
    :raises FileNotFoundError: If the file at path can't be found.
    """
    with open(path) as fp:
        return fp.read().strip()
```
`.strip()` without an arg is better here, since it cleans all whitespace chars

Luc1412 (changes requested):
Nice addition. I improved the code and updated the naming to adapt to common naming convention of these ENV vars

Luc1412 (approved):

PR #885, opened 2025-11-12
chore: remove line numbers from translations location markers

Luc1412:
This PR removes the line numbers from translation files to reduce diffs


aleyna72072 (approved):
Tried both new translate commands, works well! 🎉

PR #891, opened 2025-11-14
manage: fix `config export full`

aleyna72072:
As the default mode for open is "r", it can't create any file right now. Also adds descriptions to `sudo edumfa-manage config import -h` and `sudo edumfa-manage config export -h`


aleyna72072:
Added to milestone as the feature is broken otherwise and the fix is small.

Luc1412 (approved):

PR #894, opened 2025-11-18
set permissions when creating files not after

aleyna72072:
Sets the permission when creating the enkey, the audit keys, backups, full config backups and also the private key of the CA connector.
This could prevent an attacker racing to read the file before the permissions are set - especially the DB dump which is unprotected while the tar command is running (which could take a long time).  
In case of a full export with passwords, this could even mean that any user could read resolver passwords.

### DIFF
--- a/edumfa/commands/manage/backup.py
+++ b/edumfa/commands/manage/backup.py
@@ -45,7 +45,7 @@ def _write_mysql_defaults(filename, username, password):
     """
     Write the defaults_file for mysql commands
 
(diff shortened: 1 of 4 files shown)


aleyna72072:
Adding to milestone since this is a security issue. Or is this breaking backwards compatibility?

aleyna72072:
~Drafting since unit tests are failing.~

fritterhoff (approved):

PR #901, opened 2025-11-21
fix: don't stamp and provide output for user

### LINKED ISSUE #881: image stamps unconditionally
The stamping introduced in #866 always stamps even if there is a stamp already given. This causes the DB version to get updated without the migrations actually being run.

fritterhoff:
Instead of stamping during create_tables we go back to the previous behavior and use the old check but provide the stderror for the user.

fixes #881

### DIFF
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -15,7 +15,7 @@ edumfa-manage -q create_audit_keys || true
 echo "Creating DB"
 # FIXME: this creates a exception trace on every attempt
 attempts=10
-until edumfa-manage -q create_tables --stamp
+until edumfa-manage -q create_tables 
 do
   if [[ $attempts -eq 0 ]]; then
     echo "Exhausted database connection tries. Stopping."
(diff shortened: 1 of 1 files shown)


aleyna72072 (approved):
Tested it and LGTM. Since we have `set -e` it will exit right after asking for the current revision. Without `-e`, it'd try to stamp and fail, so since it's likely a rare case this is fine imo:

```
$ STAMP=$(sudo edumfa-manage -q db current -d /opt/edumfa/lib/edumfa/migrations/)
Can not import grpc modules.
ERROR [flask_migrate] Error: Can't locate revision identified by 'iamnotarevision'
ERROR [flask_migrate] Error: Can't locate revision identified by 'iamnotarevision'
$ if [[ -z "${STAMP//Running online/}" ]]; then echo "Stamp!"; fi
Stamp!
$ sudo edumfa-manage -q db stamp head -d /usr/local/lib/edumfa/migrations 
Can not import grpc modules.
$ echo $?
1
$ echo 'SELECT * FROM alembic_version;' | sudo mysql edumfa
version_num
iamnotarevision
```

PR #906, opened 2025-11-27
feat: make resolving user optional

### LINKED ISSUE #907: Make resolving users optional when getting token lists via API

What are you trying to achieve?

Asking LDAP or external DB for every single uid slows down fetching large token lists considerably.

In a demo setup it takes about 30 minutes to fetch a list of around 100.000 tokens via REST API because of the LDAP requests generated.

The user_id and user realm from eduMFAs DB are sufficient in many cases.

The existing PR introduces an optional parameter "resolve_users" to the token list API, which can be set to "False" to disable resolving users. The param defaults to "true", which keeps the existing behavior.

# Implementation

PR: #906

j-hoff:
This makes resolving users via LDAP or SQL optional when getting a token list.
Asking LDAP for every single uid slows down fetching large token lists considerably.
This introduces an optional parameter "resolve_users" to the token list API, which can be set to "False" to disable resolving users. Param defaults to true which keeps the existing behavior.


fritterhoff (commented):
Shouldn't it even be possible to use the userid directly from the DB?

j-hoff:
> Shouldn't it even be possible to use the userid directly from the DB?

Exactly, that is what we want to do. 
Currently, in the token response there are the following properties related to the user:

e.g.
``` 
 "user_id": "uid=someuser,ou=users,o=someorg,c=de",
 "user_realm": "someorg.de",
 "user_editable": false,
 "username": "someuser"
```
where `user_id` and `user_realm` in the token response just contain the values from the DB.
If resolving users is active (current and default behaviour) for each token the `user_id` is looked up in LDAP (or other resolver) and a user object is created. the `username` is looked up in the resolver.
With this PR, if `resolve_users` is set to `false` in the request, no lookup is made and only the first two properties are included:
``` 
 "user_id": "uid=someuser,ou=users,o=someorg,c=de",
 "user_realm": "someorg.de"
```
(We could even include `user_editable` if needed in any case.)
I made this an option, so it does not break functionality and a lookup might be desired in some cases (to check if user is active?, or `user_id` does not contain the username).

fritterhoff (approved):

PR #921, opened 2025-12-09
fix: correct parsing username from dropdown with parentheses in givenName

aleyna72072:
If the givenName contains left parenthesis, the greedy `.*` will match too much. Instead do it lazily so it stops at the first.


aleyna72072:
I think it'd be nice to have this in 2.9.0 if it was possible.

fritterhoff (approved):

PR #925, opened 2025-12-12
fix: prevent loading of inactive passkey

fritterhoff:
Right now it is possible to (partially) authenticate with a passkey of an inactive user. 
The response does not contain any user information and thus gets aborted by e.g. FUDIS but throws an error here.

This fix catches this corner case in eduMFA and aborts the authenication here.


Luc1412 (commented):
Looks good, but CI needs to be fixed

Luc1412 (approved):

PR #929, opened 2025-12-18
add bash completion support to ubuntu packages

aleyna72072:
Adds simple bash completion to the Ubuntu packages. Drafted because:  
- slow: around 1s per completion
- tab has to be pressed twice?
- TODO protobuf has to be installed
- TODO test .deb with 2.9.0 release
- TODO testing of all of the sub-commands

Theoretically it could at some point quite easily also auto-complete e.g. policy names. The problem is that the user in question would have to have read access to the config file, which is likely only the case as root or the edumfa user.

### DIFF
--- a/deploy/ubuntu-server/edumfa-apache2.postinst
+++ b/deploy/ubuntu-server/edumfa-apache2.postinst
@@ -187,6 +187,10 @@ set_path() {
   ln -sf /opt/edumfa/bin/edumfa-token-janitor /usr/bin/
 }
 
+install_completions() {
+  _EDUMFA_MANAGE_COMPLETE=bash_source  edumfa-manage > /etc/bash_completion.d/edumfa-manage-prompt
+}
+
 case "$1" in
   configure)
     export PATH=$PATH:/opt/edumfa/bin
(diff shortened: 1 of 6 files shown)


aleyna72072:
currently:
```
$ time COMP_WORDS="${COMP_WORDS[*]}" COMP_CWORD=0 _EDUMFA_MANAGE_COMPLETE=bash_complete edumfa-manage > /dev/null
Can not import grpc modules - No module named 'google.protobuf'

real    0m1,341s
user    0m1,217s
sys     0m0,112s
```
With Python 3.15.0a5 and lazy imports (`-X lazy_imports`):
```
time COMP_WORDS="${COMP_WORDS[*]}" COMP_CWORD=0 _EDUMFA_MANAGE_COMPLETE=bash_complete edumfa-manage > /dev/null
Can not import grpc modules - No module named 'google.protobuf'

real    0m1,347s
user    0m1,209s
sys     0m0,125s
```

Maybe retry when 3.15 is releases. If it still doesn't improve performance, then some kind of static expansion seems to make more sense.

aleyna72072:
WIP for dumb completion: https://gist.github.com/aleyna72072/854a83654f36fadbcd01e72525ffc21c  
For path I could just reuse the codeword.

PR #936, opened 2025-12-19
chore: prepare release v2.9.0rc1

aleyna72072:
Issues with 2.9.0:
- documentation is missing some new information: https://github.com/eduMFA/eduMFA/pull/940
- edumfa.cfg gets appended with unnecessary comment for Ubuntu packages https://github.com/eduMFA/eduMFA/pull/939  

Other than that it seems to work pretty nicely so far.
I've chosen "2.9.0rc1" instead of the format recommended by semver.org (2.9.0-rc.1), as this is explicitly listed a compliant example: https://packaging.python.org/en/latest/specifications/version-specifiers/#examples-of-compliant-version-schemes  
The Debian version numbers are taken from here: https://wiki.debian.org/Versioning. They are different but should guarantee that they are ordered earlier than 2.9.0noble. I am unsure if this would be the case with 2.9.0rc1.

### DIFF
--- a/READ_BEFORE_UPDATE.md
+++ b/READ_BEFORE_UPDATE.md
@@ -33,6 +33,100 @@
         * Replace all occurrences of `privacyideaserver_write` in `policy.action` with `edumfaserver_write`
         * Replace all occurrences of `privacyidea.` in `smsgateway.providermodule` with `edumfa.`
 
+## eduMFA 2.9.0
+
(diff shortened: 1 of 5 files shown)


aleyna72072:
I've tested the Ubuntu package with upgrading and a fresh install, and besides those two issues above it works. It should probably not get uploaded to this Nexus repo though as it's a pre-release: https://github.com/eduMFA/eduMFA/blob/main/.github/workflows/ubuntu.yml#L61

Luc1412 (approved):

PR #945, opened 2025-12-23
feat: add option for token-janitor to remove tokeninfo

### LINKED ISSUE #764: edumfa-token-janitor - add action to remove tokeninfo-key
We are using the `edumfa-token-janitor` script to find and delete tokens which are stuck in `verify` or `clientwait` state. Basically, we check for stuck tokens and use the mark action to add a tokeninfo key `to_be_deleted` with a current timestamp. 24 hours later we delete it. If someone didn't finish the verification or webauthn enrollment by then, it's probably broken.

For example

```
/opt/edumfa/bin/edumfa-token-janitor find \
   --tokenattribute rollout_state --tokenattribute-value verify \
   --has-not-tokeninfo-key to_be_deleted --action mark \
   --set-tokeninfo-key to_be_deleted --set-tokeninfo-value $tomorrow 
```

However, if the token has been verified in the meantime we would like to remove the tokeninfo-key `to_be_deleted` again from the token. But that seems not possible. The mark action only adds tokeninfo, but cannot remove it. Setting the value to an empty string doesn't remove it either.

So an option `--remove-tokeninfo-key` would be helpful to remove the key (or `--remove-tokeninfo-key` and `--remove-tokeninfo-value` if tokeninfo can be an array of values, to remove a specific value only).

Some discussion related to this is on the mailing list in this thread:

https://www.listserv.dfn.de/sympa/arc/edumfa-users/2025-09/msg00001.html

pmainz:
This adds the possibility to remove a former set tokeninfo-attribute to the token-janitor.

### DIFF
--- a/tools/edumfa-token-janitor
+++ b/tools/edumfa-token-janitor
@@ -815,13 +815,18 @@ def find(
                                 token_obj.save()
                         elif action == "unmark":
                             if remove_tokeninfo_key:
-                                click.echo(
-                                    f"Removing tokeninfo for token {token_obj.token.serial!s}: {remove_tokeninfo_key!s}"
-                                )
-                                token_obj.del_tokeninfo(
-                                    remove_tokeninfo_key
-                                )
-                                token_obj.save()
+                                if not token_obj.get_tokeninfo(key=remove_tokeninfo_key):
+                                    click.echo(
+                                        f"Could not find tokeninfo-key {remove_tokeninfo_key!s} on token {token_obj.token.serial!s}"
+                                    )
+                                else:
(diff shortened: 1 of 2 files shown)


j-hoff (approved):

fritterhoff (approved):

tools/edumfa-token-janitor
fritterhoff, line 93: Just a little nit-picky. We could add the unmark action also here ;-)

pmainz: uuh... yes! I have overlooked that.

Issue #920, opened 2025-12-09
Debian 13 with MariaDB: /validate/check sometimes fails due to "Record has changed since last read in table"


aleyna72072:
On Debian 13 with a MariaDB eduMFA sometimes returns 500 on `/validate/check`. The log files say updating the `clientapplication` or `token` table failed due to "Record has changed since last read in table".

### Reproduce

Run 2 requests at the same time, e.g.: `curl -k 'https://localhost/validate/check?user=username&pass=pwd123456' & curl -k 'https://localhost/validate/check?user=username&pass=pwd123456'`  
One suceeeds, the other one produces the error below.

### Potential explanation

Debian 13 currently ships mariadb-server 11.8.3. Starting 11.6.2, `innodb_snapshot_isolation` [has a new default of ON](https://mariadb.com/docs/server/server-usage/storage-engines/innodb/innodb-system-variables#innodb_snapshot_isolation).  
The explanation is likely in [the blog post](https://mariadb.com/resources/blog/isolation-level-violation-testing-and-debugging-in-mariadb/), but the details are still unclear to me.

### Workaround

- Setting `innodb_snapshot_isolation` to `OFF` _might_ revert to the old behaviour. This is **strongly discouraged** as this might open up reusing of OTPs!
- Setting `transaction-isolation          = READ-COMMITTED` avoids this issue, too. This is not recommended, as the consequences of changing away from the default `REPEATABLE-READ` are currently unknown.
  + In the [commit](https://github.com/MariaDB/server/commit/b8a671988954870b7db22e20d1a1409fd40f8e3d) it says "Locking reads inside InnoDB used to read the latest committed version" which makes me lead to believe that in this regard it could behave similar?

### Open Questions

- How could a proper fix look like which avoids changing MariaDB defaults?
- Is the old behaviour causing practical problems and which?

### Configuration

* **eduMFA version**: 2.8.0
* **Installation method**: PyPI
* **Python version**: 3.13
* **Operating system**: Debian 13
* **Webserver**: Apache2
* **Token database**: MariaDB 1:11.8.3-0+deb13u1

### Log file

```
[2025-12-04 16:26:55,128][1569911][140579844335296][ERROR][edumfa.app:875] Exception on /validate/check [POST]
Traceback (most recent call last):
  File "/opt/edumfa/lib/python3.13/site-packages/sqlalchemy/engine/base.py", line 1964, in _exec_single_context
    self.dialect.do_execute(
[...]

PR #958, opened 2026-01-09
fix: try handling the database errors

fritterhoff:
Just as a first attempt to handle the error(s) that are mentioned in #920

### DIFF
--- a/edumfa/api/before_after.py
+++ b/edumfa/api/before_after.py
@@ -32,6 +32,7 @@ import logging
 import threading
 
 from flask import current_app, g, request
+from sqlalchemy.exc import IntegrityError, OperationalError
 
 from edumfa.api.auth import admin_required, jwtauth, user_required
 from edumfa.api.lib.postpolicy import postrequest, sign_response
@@ -84,8 +85,6 @@ from .system import system_blueprint
 from .token import token_blueprint
 from .tokengroup import tokengroup_blueprint
 from .user import user_blueprint
-
-from sqlalchemy.exc import OperationalError, IntegrityError
 
 log = logging.getLogger(__name__)
 
--- a/edumfa/api/before_after.py
+++ b/edumfa/api/before_after.py
@@ -427,7 +427,7 @@ def internal_error(error):
 @validate_blueprint.app_errorhandler(IntegrityError)
 def sql_error(error):
     """
-    This function is called when an database error occurs.
+    This function is called when a database error occurs.
     """
     log.error("Database error occurred: {!r}".format(error))
     if "audit_object" in g:
--- a/edumfa/api/before_after.py
+++ b/edumfa/api/before_after.py
@@ -85,6 +85,8 @@ from .token import token_blueprint
 from .tokengroup import tokengroup_blueprint
 from .user import user_blueprint
 
+from sqlalchemy.exc import OperationalError, IntegrityError
+
 log = logging.getLogger(__name__)
 
 
@@ -421,3 +423,14 @@ def internal_error(error):
     if "audit_object" in g:
         g.audit_object.log({"info": str(error)})
     return send_error(str(error), error_code=-500), 500
+
+@validate_blueprint.app_errorhandler(OperationalError)
+@validate_blueprint.app_errorhandler(IntegrityError)
+def sql_error(error):
+    """
+    This function is called when an database error occurs.
+    """
+    log.error("Database error occurred: {!r}".format(error))
+    if "audit_object" in g:
+        g.audit_object.log({"info": "Database error occurred."})
+    return send_error("A database error occurred.", error_code=-600), 500
--- a/tests/test_api_validate.py
+++ b/tests/test_api_validate.py
@@ -11,6 +11,7 @@ from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import serialization
 from cryptography.hazmat.primitives.asymmetric import rsa
 from passlib.hash import argon2
+from pymysql import OperationalError as PyMysqlOperationalError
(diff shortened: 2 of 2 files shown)


edumfa/api/before_after.py
```diff
@@ -421,3 +422,14 @@ def internal_error(error):
     if "audit_object" in g:
         g.audit_object.log({"info": str(error)})
     return send_error(str(error), error_code=-500), 500
+
+@validate_blueprint.app_errorhandler(OperationalError)
+@validate_blueprint.app_errorhandler(IntegrityError)
+def sql_error(error):
+    """
+    This function is called when an database error occurs.
```
aleyna72072: an -> a

aleyna72072:
Maybe add this to also check the log message in the unit test:

```
From c9f1f0a49fca08d12309587dbf10af875ae2e35b Mon Sep 17 00:00:00 2001
From: Aleyna Nasher Taher <aleyna.nashertaher@daasi.de>
Date: Mon, 12 Jan 2026 13:24:00 +0100
Subject: [PATCH] SQUASHME: test_20_db_error check log message

---
 tests/test_api_validate.py | 26 +++++++++++++++++++-------
 1 file changed, 19 insertions(+), 7 deletions(-)

diff --git a/tests/test_api_validate.py b/tests/test_api_validate.py
index bc0635bf3..4d4a0a2ec 100644
--- a/tests/test_api_validate.py
+++ b/tests/test_api_validate.py
@@ -4,6 +4,7 @@ import json
 import logging
 import time
 from urllib.parse import quote, urlencode
+from pymysql import OperationalError as PymysqlOperationalError
 
 import mock
 import responses
@@ -6539,7 +6540,8 @@ class AChallengeResponse(MyApiTestCase):
         remove_token("hotp_serial")
         delete_policy("increase_failcounter_on_challenge")
 
-    def test_20_db_error(self):
+    @log_capture(level=logging.DEBUG)
+    def test_20_db_error(self, capture):
         # Create HOTP token
         init_token(
             {
@@ -6551,9 +6553,9 @@ class AChallengeResponse(MyApiTestCase):
             user=User("cornelius", self.realm1),
         )
         with self.app.test_request_context(
-                "/validate/check",
-                method="POST",
-                data={"user": "cornelius", "pass": "pin"},
+            "/validate/check",
+            method="POST",
+            data={"user": "cornelius", "pass": "pin"},
         ):
             res = self.app.full_dispatch_request()
             self.assertEqual(200, res.status_code)
@@ -6563,20 +6565,30 @@ class AChallengeResponse(MyApiTestCase):
             from sqlalchemy.exc import OperationalError
 
             mock_check_token_list.side_effect = OperationalError(
-                statement=None, params=None, orig=None
+                statement=None, params=None, orig=PymysqlOperationalError("TestMessage")
             )
 
             with self.app.test_request_context(
                 "/validate/check",
                 method="POST",
-                data={"user": "cornelius", "pass": "pin"},
+                data={"user": "cornelius", "pass": "secret_password123"},
             ):
                 res = self.app.full_dispatch_request()
                 self.assertEqual(500, res.status_code)
                 result = res.json.get("result")
                 self.assertFalse(result.get("status"))
-                self.assertEqual("A database error occurred.", result.get("error").get("message"))
+                self.assertEqual(
+                    "A database error occurred.", result.get("error").get("message")
+                )
                 self.assertEqual(-600, result.get("error").get("code"))
+                log_messages = str(capture)
+                self.assertNotIn("secret_password123", log_messages, log_messages)
+                self.assertIn(
+                    "Database error occurred: OperationalError('(pymysql.err.OperationalError) TestMessage')",
+                    log_messages,
+                    log_messages,
+                )
+
 
 class TriggeredPoliciesTestCase(MyApiTestCase):
     def setUp(self):
-- 
2.52.0


```

aleyna72072:
Do you think this could also replace https://github.com/eduMFA/eduMFA/commit/d4cab210728ac8c7c908c934639576b3e61230ac ? It seems to work the same in my partial testing.  

EDIT: this would cancel the auth flow so we keep it - @fritterhoff via Matrix

aleyna72072 (changes requested):
Everything I tried seems to work, and it looks like an elegant way to solve it systematically.  
Would be interested to know your opinion on the two comments right above, though.

tests/test_api_validate.py
```diff
@@ -6539,6 +6540,53 @@ def test_19_increase_failcounter_on_challenge(self):
...
+            from sqlalchemy.exc import OperationalError
+
+            mock_check_token_list.side_effect = OperationalError(
+                statement=None, params=None, orig=PyMysqlOperationalError("TestMessage")
+            )
+
+            with self.app.test_request_context(
+                "/validate/check",
+                method="POST",
+                data={"user": "cornelius", "pass": "pin"},
```
aleyna72072, line 6574: You removed the check for the password not being in the log message, was this on purpose?

fritterhoff: From my point of view this is not changed/covered by this change. 
If we want to ensure that the pin is not logged I would suggest a new test.

aleyna72072: My reason for adding that was that we directly log a DB exception, do we do this somewhere else already? If not, I'd argue this is affected by this change?

aleyna72072: But if you say this is fine I'm inclined to trust you :smile: .

fritterhoff: I think it would make sense if the pin was logged before? Right now I can't find the pin in the stacktrace so that is nothing that is fixed by this PR and I would keep it of of scope for this PR.

aleyna72072 (approved):
LGTM

PR #975, opened 2026-01-18
feat: optimize the logging of the docker entrypoint regarding the password

fritterhoff:
Until now the password is always printed to the stdout.

From my point of view this only makes sense if the password is not user defined and generated by the script.

### DIFF
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -3,6 +3,12 @@ set -e
 
 GEN_PWD="$(openssl rand -base64 42)"
 EDUMFA_ADMIN_USER="${EDUMFA_ADMIN_USER:-admin}"
+# Check if password is set, otherwise generate one and store the information if geneated in a variable
+GENERATED_PASSWORD=0
+if [ -z "$EDUMFA_ADMIN_PASS" ]; then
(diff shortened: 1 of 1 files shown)


deploy/docker/entrypoint.sh
```diff
@@ -3,6 +3,12 @@ set -e
 
 GEN_PWD="$(openssl rand -base64 42)"
 EDUMFA_ADMIN_USER="${EDUMFA_ADMIN_USER:-admin}"
+# Check if password is set, otherwise generate one and store the information if geneated in a variable
```
aleyna72072:
```suggestion
# Check if password is set, otherwise generate one later.
```
typo

deploy/docker/entrypoint.sh
```diff
@@ -3,6 +3,12 @@ set -e
...
+GENERATED_PASSWORD=0
+if [ -z "$EDUMFA_ADMIN_PASS" ]; then
+  echo "No EDUMFA_ADMIN_PASS set, generating a random one and printing it afterwards."
```
aleyna72072:
```suggestion
  echo "No EDUMFA_ADMIN_PASS set, a random password will be generated and printed when initialization finishes."
```
To avoid users being confused why it says "afterwards" but no credentials follow.

aleyna72072 (approved):
LGTM

PR #977, opened 2026-01-21
fix: correct undeletable token and janitor orphan search for `PasswdIdResolver`

aleyna72072:
Currently it throws an Exception "'NoneType' object is not subscriptable". This makes deleting orphaned tokens of PasswdIdResolver in the webui impossible. Also they don't show up in the token janitor's orphan search.  

Also removes the notice about Python2 encoding difficulties as its not relevant for Python3.

### DIFF
--- a/edumfa/lib/resolvers/PasswdIdResolver.py
+++ b/edumfa/lib/resolvers/PasswdIdResolver.py
@@ -247,17 +247,20 @@ class IdResolver(UserIdResolver):
 
         return ret
 
-    def getUsername(self, userId):
+    def getUsername(self, userId: str) -> str:
         """
-        Returns the username/loginname for a given userid
-        :param userid: The userid in this resolver
-        :type userid: string
-        :return: username
-        :rtype: str
+        Returns the username/loginname for a given userid, or an empty string
+        if it can't be found.
+
+        :param userid: the userid of the user in this resolver
+        :return: username/loginname of the userid
         """
         fields = self.descDict.get(userId)
-        index = self.sF["username"]
-        return fields[index]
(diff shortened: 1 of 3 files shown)


aleyna72072:
FYI: The LDAP resolver returns an empty string for the same method for that case, so I just copied that.

aleyna72072:
@Luc1412 As suggested, I've split the deprecation fixes (#979) and the removal of possibly inappropriate wording into their own PRs (#980).

edumfa/lib/resolvers/PasswdIdResolver.py
```diff
@@ -254,15 +256,20 @@ def getUserInfo(self, userId, no_passwd=False):
 
     def getUsername(self, userId):
         """
-        Returns the username/loginname for a given userid
+        Returns the username/loginname for a given userid, or an empty string
+        if it can't be found.
```
Luc1412, line 253: I believe returning None and handling that would be better. Additionally we could start adding type hints if we know the types

Luc1412 (changes requested):
Merges Conflicts need to be resolved and I got an annotation regarding the meth signature.

edumfa/lib/resolvers/PasswdIdResolver.py
aleyna72072: Returning an empty string is how the other resolvers do this; changing that is out of scope for this fix imo.  

> Additionally we could start adding type hints if we know the types 

Sounds good, I've added one for the getUsername function there.

Luc1412 (approved):

PR #978, opened 2026-01-23
fix: use correct variable for webauthn reset

fritterhoff:
Right now a undefined variable is used. Prop this will break the flow.
No idea why I did not notice during testing...


aleyna72072:
If the plan is to keep that code around longer: are there unittests for the original function you could port to this?

fritterhoff:
Unit tests for this could be added in the future but would require simulating a complete passkey setup in combination with multiple policies.

PR #979, opened 2026-01-23
fix some deprecations

### LINKED ISSUE #976: python 3.14 compat: edumfa-pip-update uses removed library
> Deprecated since version 3.11, removed in version 3.13.
> Applications should use the [subprocess](https://docs.python.org/3/library/subprocess.html#module-subprocess) module instead.

https://docs.python.org/3/library/pipes.html

https://github.com/eduMFA/eduMFA/blob/v2.9.0/tools/edumfa-pip-update#L37

Unrelated notes:
- dead link on https://edumfa.readthedocs.io/en/v2.9.0/installation/upgrade.html "READ_BEFORE_UPDATE"
- maybe check if the script can warn if update is unavailable due to the python version

aleyna72072:
While going through some unit tests, I noticed small deprecations. This patch fixes them before they might cause issues.  

fixes #976

### DIFF
--- a/edumfa/api/lib/utils.py
+++ b/edumfa/api/lib/utils.py
@@ -44,7 +44,6 @@ from ...lib.error import ERROR, AuthError, ParameterError
 from ...lib.log import log_with
 
 log = logging.getLogger(__name__)
-ENCODING = "utf-8"
 TRUSTED_JWT_ALGOS = [
     "ES256",
     "ES384",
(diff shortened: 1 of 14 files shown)


Luc1412 (commented):
Looks mostly good, but whats the reasoning behind the addition of `True` within `and_`? Also merge conflicts need to be resolved.

aleyna72072:
> Looks mostly good, but whats the reasoning behind the addition of `True` within `and_`?

As per the commit description:
> SADeprecationWarning: Invoking and_() without arguments is deprecated, and will
be disallowed in a future release.   For an empty and_() construct, use
'and_(true(), *args)' or 'and_(True, *args)'.

Same goes for or_.

> Also merge conflicts need to be resolved.

Done.

Luc1412 (approved):

PR #995, opened 2026-02-06
chore: fix typos

Luc1412:
This PR is a WIP of fixing some typos.


edumfa/lib/security/default.py
```diff
@@ -257,10 +257,10 @@ def __init__(self, config=None):
             cipher = f.read()
 
         if len(cipher) > 100:
-            config["crypted"] = True
+            config["encrypted"] = True
 
-        if is_true(config.get("crypted", "")):
-            self.crypted = True
+        if is_true(config.get("encrypted", "")):
```
aleyna72072: This sounds like it  might break existing setups? Did not check what the config option is though. If so, I'd imagine we need to wait until 3.0.0 to change this.

tests/test_lib_crypto.py
```diff
@@ -76,7 +76,7 @@ def test_01_no_file_in_config(self):
     def test_04_random(self):
         config = current_app.config
         hsm = DefaultSecurityModule(
-            {"file": config.get("EDUMFA_ENCFILE"), "crypted": True}
+            {"file": config.get("EDUMFA_ENCFILE"), "encrypted": True}
```
aleyna72072: see above, maybe this test can actually help?

tests/test_lib_crypto.py
```diff
@@ -144,9 +144,9 @@ def test_06_password_encrypt_decrypt(self):
     def test_07_encrypted_key_file(self):
         config = current_app.config
         hsm = DefaultSecurityModule(
-            {"file": config.get("EDUMFA_ENCFILE_ENC"), "crypted": True}
+            {"file": config.get("EDUMFA_ENCFILE_ENC"), "encrypted": True}
```
aleyna72072: see above

tools/edumfa-expired-users
```diff
@@ -149,7 +149,7 @@ def expire(realm, attribute_name, delete_serial, unassign_serial, noaction):
                 # 0 -> account never expires
                 print(f"{username} does not expire.")
                 continue
-            td = datetime.timedelta(seconds=(int(account_expires) / MULTIPLYER))
+            td = datetime.timedelta(seconds=(int(account_expires) / MULTIPLIER))
```
aleyna72072, line 152: Not sure what to do about this, but those variables aren't set: MULTIPLIER, MS_AD_START  
This is not a result of this PR so you can obviously ignore it, but maybe you have a quick fix at hand.

aleyna72072 (changes requested):
Big fan of your systematic clean-up of the code base!  
The only concern I have is the `config.get`-stuff.

tools/edumfa-expired-users
aleyna72072: Ah, it's probably these vars: https://github.com/eduMFA/eduMFA/blob/v2.9.0/edumfa/lib/resolvers/LDAPIdResolver.py#L86-L88

tools/edumfa-expired-users
aleyna72072, line 70:
```suggestion
expect.

FIXME: This script seems to be broken as it contains undefined variables      
(MS_AD_START and MULTIPLIER). The source for these constants is probably:      
https://github.com/eduMFA/eduMFA/blob/v2.9.0/edumfa/lib/resolvers/LDAPIdResolver.py#L86-L88
```

tools/edumfa-expired-users
aleyna72072: Added a suggestion to document these at least.

tools/edumfa-expired-users
Luc1412: Think this is out of scope of this PR. Could you create an issue instead?

aleyna72072: https://github.com/eduMFA/eduMFA/issues/997

aleyna72072 (approved):
LGTM

PR #998, opened 2026-02-09
chore: migrate development enviroment to uv

Luc1412:
This PR migrates and requires the development enviroment enviroment to use UV.

## Direct dependencies are now defined in `pyproject.toml`
Before all direct and transitive dependencies where pinned in a lock-like requirements.txt file. This caused all installations to be deterministic and may caused conflicts with dependencies unrelated to eduMFA
With this change all direct depedencies are now pinned in the `pyproject.toml`, so transitive dependency versions are more flexible.
## Dev dependencies are now in a dependency group
Before dev dependencies (lint, test, docs) where optional dependencies. Thus they where installable with `pip install edumfa[test]` or  `pip install .[test]` when installing from source. This no longer work, since they got moved into dependency groups. This allows to install them with `uv sync` in a dev env or `uv sync --group tests` in a prod env.
## Updated deterministic installation dependencies
The `requirements.txt` has been updated for a better deterministic installation. Dependencies that are only required for specific python versions or OS's will now only be installed if required. Also missing transitive dependencies where added


README.rst
```diff
@@ -63,20 +63,9 @@ Setup
 For setting up the system to *run* it, please read install instructions 
 at `edumfa.readthedocs.io <http://edumfa.readthedocs.io/en/latest/installation/index.html>`_.
 
-If you want to setup a development environment start like this::
+If you want to setup a development, you need to have (UV)[https://docs.astral.sh/uv/getting-started/installation/] installed. Then you can run::
```
aleyna72072:
```suggestion
If you want to setup a development, you need to have [UV](https://docs.astral.sh/uv/getting-started/installation/) installed. Then you can run::
```

aleyna72072 (changes requested):
The makefiles
- /Makefile
- /doc/Makefile

still directly run the tools (sphinx, pybabel). To my understanding, they should instead be run with `uv run`?  
Also it'd be nice if the Github build action actually used the makefile if it is possible, so we don't have two build instructions.

README.rst
```diff
@@ -63,20 +63,9 @@ Setup
...
-    source venv/bin/activate
-    pip install .
-    
-.. _testing_env:
-
-You may additionally want to set up your environment for testing, by adding the
-additional dependencies::
-
-    pip install .[test]
+    uv sync
```
aleyna72072, line 68:
```suggestion
    uv sync --frozen
```
**This is only a shot in the dark**, but currently running `uv sync` causes the lock file to be changed, which makes it annoying to jump between branches. I also assume this can cause commits accidentally containing changes to the lock files.  
As far as I understand, we only want to install the dependencies and not update the lockfile here. Do you think this is a good way to solve this?

CONTRIBUTING.md
```diff
@@ -31,7 +31,7 @@ https://github.com/eduMFA/eduMFA/issues
 The source code is pretty well documented. The main documentation resides in the
 `doc`-subfolder and can be build with:
 ```
-pip install .[docs]
+uv sync
```
aleyna72072, line 34:
```suggestion
uv sync --frozen
```
See README comment, but again this is just me guessing.

Luc1412:
> The makefiles
> 
> * /Makefile
> * /doc/Makefile
> 
> still directly run the tools (sphinx, pybabel). To my understanding, they should instead be run with `uv run`? Also it'd be nice if the Github build action actually used the makefile if it is possible, so we don't have two build instructions.

Done. Also switched CI to use the makefile

README.rst
Luc1412: I think it's not needed. The lockfile shouldn't change, as long you don't update the dependencies. Every branch also got their own lockfile.

aleyna72072: Ah, so this change being applied on every `uv sync` was because of something else? https://github.com/eduMFA/eduMFA/pull/998/changes/54881f9cdda43df31d393b6e454c713d26a50bf5

Luc1412: This was because of me updating the pyproject.toml but not updating the lock.

aleyna72072: Ah, thanks for explaining! Would it make sense to have a CI check for this like for the translations? To avoid a similar situation.

Luc1412: I don't think this is needed. Dependencies are usually changed by renovate.

aleyna72072: Alright, please re-request my review as soon as you consider it ready :slightly_smiling_face: .

aleyna72072 (approved):
LGTM

PR #1002, opened 2026-02-11
chore: replace pytz with standard library zoneinfo

Luc1412:
Standard lib also provides a utc timezone object, so we don't need pytz here. 

This is in preperation to remove pytz as a direct dependency in #998


aleyna72072 (approved):
LGTM

PR #1007, opened 2026-02-15
feat: add missing functionality to enable tokens again

### LINKED ISSUE #1004: action token janitor: enable
We disable (and mark "unused: $date") token without user (--orphaned 1).
After some days we delete them finally.
Sometimes user will be re-actived, multiple reasons possible, but with edumfa-token-janitor we can't activate the token. There is no action within edumfa-token-janitor. Or did I oversee something?

edumfa-token-janitor action enable

fritterhoff:
This pull request adds support for an `enable` action to the `edumfa-token-janitor` tool, making it possible to enable previously disabled tokens. The new action is integrated alongside the existing actions and is reflected in both the allowed actions list and the user documentation.

New token action support:

* Added `"enable"` to the `ALLOWED_ACTIONS` list in `tools/edumfa-token-janitor`, allowing the tool to recognize and process the new action.
* Updated the documentation and help text to include the `enable` action in the list of available actions.
* Implemented the logic for the `enable` action in the `find` function, calling `enable_token` with `enable=True` and providing user feedback when a token is enabled.

### DIFF
--- a/tools/edumfa-token-janitor
+++ b/tools/edumfa-token-janitor
@@ -70,6 +70,7 @@ ALLOWED_ACTIONS = [
     "export",
     "listuser",
     "tokenrealms",
+    "enable",
 ]
 
 __doc__ = f"""
@@ -93,6 +94,7 @@ Actions:
 * unmark
 * disable
 * delete
+* enable
 
 
     edumfa-token-janitor find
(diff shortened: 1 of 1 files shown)


tools/edumfa-token-janitor
j-hoff: It can list, enable, disable, delete, mark, unmark or unassign tokens

j-hoff (approved):
LGTM, we should probably update the docs as well?

PR #1010, opened 2026-02-23
fix(webui): token wizard silently ignoring default tokentype

aleyna72072:
Currently, when a user has no permission to enroll the default tokentype, the UI presents the first enrollable type instead. This PR adds a warning so administrators can at least know why this is happening.  
This warning would not show if there is no default tokentype set, as this would suddenly make a warning appear on existing configurations since a default_tokentype is technically not required for the token wizard.  

grafik


aleyna72072:
Added information why hotp is explicitly excluded.

Luc1412 (approved):

Issue #985, opened 2026-01-25
docker: mapping `/etc/edumfa/` prevents updates in `edumfa.py` or `logging.yml`


fritterhoff:
## Summary

Right now we are mapping `/etc/edumfa` to a docker volume here:
https://github.com/eduMFA/eduMFA/blob/93fac4bf8a2e142de326f1d29419698c514a2385/deploy/docker-example/docker-compose.yml#L25

In case we update the `edumfa.py` or the `logging.yml` these files don't get updated in a running instance.

## Reproduce

- Fire up a docker compose setup
- Update the `logging.yml` or the `edumfa.py` inside the repo
- Rebuild the image
- Run another `docker compose up`
- Changes don't get applied, since the old files are mounted to the existing docker volume

## Ideas

- Either read the keys separately
- Place the `edumfa.py` and `logging.yml` somewhere else
- Place the keys in a subfolder.

aleyna72072:
> Place the keys in a subfolder

This is how we do it internally, at `/etc/edumfa/keys`. In the next sprint I'll try to implement it here too, if you agree.

aleyna72072:
Thinking about it, FHS says for `/etc:`

> Host-specific system configuration

https://www.pathname.com/fhs/pub/fhs-2.3.html#ETCHOSTSPECIFICSYSTEMCONFIGURATION

🤔

tribut:
The fact the default config files live in `/etc/edumfa` makes it also unnecessarily hard to use a bind-mount with env-based configuration (because the default files have to be copied manually).

This sounds like a (necessary) breaking change, though. I don't see how either moving the config or the keys can be done without breaking some existing setups. But maybe just a fatal error in `entrypoint.sh` if `/etc/edumfa/$moved_file` exists is enough to not make this dangerous.

PR #1042, opened 2026-03-07
feat(image): Allow customization options to be set using env variables

tribut:
This adds the configuration variables from https://edumfa.readthedocs.io/en/v2.9.0/faq/customization.html to the possible values set from the docker environment.

The commit also slightly changes the semantics of `get_var`: Where before, a falsy value was treated as "variable not set", it now just checks if the variable is not set (as documented).

For existing users, this somewhat depends on https://github.com/eduMFA/eduMFA/issues/985


aleyna72072 (approved):
LGTM, thanks!

PR #1055, opened 2026-03-11
docs: add integration chapter

aleyna72072:
Add a chapter about connection possibilities to the docs. Main reason is a complaint I got about there being no instructions for manual installation of RADIUS.  
The other pages are pretty bare-bones but probably enough for discoverability?

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -47,6 +47,7 @@ Table of Contents
    eventhandler/index
    periodictask/index
    audit/index
+   integrations/index
    machines/index
    workflows_and_tools/index
    jobqueue/index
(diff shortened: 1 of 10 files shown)


fritterhoff (approved):

aleyna72072:
@oculos If your plugin is ever published (under the eduMFA organization or elsewhere), ping me and I'll add it to the list :smile: .

oculos:
Do you mean the keycloak plugin or a radius one?

aleyna72072:
> Do you mean the keycloak plugin or a radius one?

The Keycloak plugin. I wasn't aware you were working on a RADIUS pluigin?

oculos:
I have a while ago modified the PrivacyIDEA plugin so that it would handle push notifications better. 

The keycloak plugin needs just to be polished when it comes to the UX for OTP registration, otherwise it is fully functional. 

Whenever the project thinks it should be added there, it’s fine by me. We will still work on it for a few months, though.

aleyna72072:
@oculos Oh, is the code for the keycloak plugin already public?

PR #1056, opened 2026-03-11
fix(image): fix EDUMFA_CONFIGFILE configs

### LINKED ISSUE #1043: image: alternative config path issue
support alternative config path via env var  
https://github.com/eduMFA/eduMFA/blob/v2.9.0/deploy/docker/entrypoint.sh#L8-L9

aleyna72072:
The check to make sure all necessary fields are set and the config is not malformed did not respect EDUMFA_CONFIGFILE. This commit fixes that.  
fixes #1043


aleyna72072:
Closed in favour of https://github.com/eduMFA/eduMFA/pull/1057

PR #1057, opened 2026-03-11
fix(image): ensure upgradability of edumfa config

### LINKED ISSUE #985: docker: mapping `/etc/edumfa/` prevents updates in `edumfa.py` or `logging.yml`

### LINKED ISSUE #1043: image: alternative config path issue

aleyna72072:
Deletes the old files if `edumfa.cfg` has the correct hash. Moves the files to `/opt`, which is not part of any volume.  
fixes #1043
fixes #985  

TODO: If merged, add a new issue for 3.0.0 to remove the workaround.

### DIFF
--- a/deploy/docker/entrypoint.sh
+++ b/deploy/docker/entrypoint.sh
@@ -1,6 +1,61 @@
 #!/usr/bin/bash
 set -e
 
+# In eduMFA v2.9.0, the config and logging file were suggested to be put into a
+# volume. This would lead to them not getting updated. This is a workaround to
+# handle this unfortunate situation:
+# If they are the same as in that release, delete them.
+# If both were changed, the operator should be aware of the situation.
+# If only one of the files was changed, the situation is ambiguous.
+# This workaround will be deleted in v3.0.0.
+config_equal="false"
+logging_equal="false"
+if [ -e "/etc/edumfa/edumfa.cfg" ]; then
+  config_hash="$(sha256sum /etc/edumfa/edumfa.cfg | cut -d " " -f 1 )"
+  if [ "$config_hash" == "a87cf5f96a5eb1ceb0f6d59c9083814e2c55bf7c3b85ca8ecdb532dc059fa064" ]; then
+    config_equal="true"
+  fi
+fi
+if [ -e "/etc/edumfa/logging.yml" ]; then
+  logging_hash="$(sha256sum /etc/edumfa/logging.yml | cut -d " " -f 1 )"
+  if [ "$logging_hash" == "447d9ab4f18505e6acc441a94a68e2a3410125c37b6bc942ad0c5523a1caf6f9" ]; then
+    logging_equal="true"
+  fi
+fi
+if [[ "$config_equal" == "true" && "$logging_equal" == "true" ]]; then
+  # No changes to either file. This should be the most common scenario, since
+  # the example compose file relied on environment variables to configure
+  # eduMFA.
+  rm -f "/etc/edumfa/edumfa.cfg"
+  rm -f "/etc/edumfa/logging.yml"
+elif [[ "$config_equal" != "true" && "$logging_equal" != "true" ]]; then
+  # Both files were edited and/or already deleted.
+  true
+else
+  # Only one file was edited, the situation is ambigious.
+  echo "=======================================================================
+  Due to an error in v2.9.0, the following files were moved:
+  /etc/edumfa/edumfa.cfg ==> /opt/edumfa/edumfa.cfg
+  /etc/edumfa/logging.yml ==> /opt/edumfa/logging.yml
+  Please update your deployments if necessary.
+  This warning will disappear in eduMFA 3.0.0."
+  echo "======================================================================="
+fi
+
+# Decide which config file to use. The hierarchy is as follows:
+# 0. Use $EDUMFA_CONFIGFILE if it is set.
+# 1. Use /etc/edumfa/edumfa.cfg if it exists.
+# 2. Use this container's /opt/edumfa/edumfa_config.py.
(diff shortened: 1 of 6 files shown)


aleyna72072:
@tribut Can't select you as a reviewer, but I would appreciate your review, too!

deploy/docker/entrypoint.sh
```diff
@@ -1,6 +1,32 @@
...
 set -e
 
+# In eduMFA v2.9.0, the config file was placed at the wrong location. If the
+# file at /etc/edumfa/edumfa.cfg has its hash, delete it to avoid it not getting
+# updated.
+if [ -e "/etc/edumfa/edumfa.cfg" ]; then
+  config_hash="$(sha256sum /etc/edumfa/edumfa.cfg | cut -d " " -f 1 )"
+  if [ "$config_hash" == "a87cf5f96a5eb1ceb0f6d59c9083814e2c55bf7c3b85ca8ecdb532dc059fa064" ]; then
+    rm -f "/etc/edumfa/edumfa.cfg"
+    # also delete logging.yml (assume it exists since the old config exists too)
```
tribut: This assumes that nobody modified `logging.yml`. If we go that route, should we not add the same checksum check / fallback machanism to `EDUMFA_LOGCONFIG`?

tribut (changes requested):
Thanks for the ping.

This PR misses one instance of `/etc/edumfa` in [`deploy/gunicorn/edumfaapp.py`](https://github.com/eduMFA/eduMFA/blob/1352a7cc09945e531ccc6cd94006eac9a9a716ee/deploy/gunicorn/edumfaapp.py#L7).

It is also a breaking change for anyone who had modified their logging config (see my comment there).

In general, I'm not sure how much I like the approach. At least I would add a big fat warning if either `edumfa.cfg` or `logging.yml` exist at the old location, so users have a chance to notice and we can one day stop calculating the checksum(s) on container startup.

My approach would have been to check if either of the files exist and if yes, print an error ("the files moved, see changelog") and exit. That way we are sure users notice (quickly) and get to fix the problem in a way that suits them.

I would be very irritated if I had mounted my `edumfa.cfg` at `/etc/edumfa` and suddenly it disappears. This would also likely cause problems when the container next starts, because when the local file does not exist, docker creates an empty file, owned by `root:root`.
EDIT: Thinking about it, you probably can't delete a mounted file from inside the container in the first place...

deploy/docker/entrypoint.sh
aleyna72072: Thanks for the extensive review! Hope you'll stick around in the community :slightly_smiling_face: .  
In general, this is a documentation compose file for configuring eduMFA with environment variables. Users of that approach would likely not edit those files, so making sure they have a trouble-free update is valuable in my eyes.

> My approach would have been to check if either of the files exist and if yes, print an error ("the files moved, see changelog") and exit.

I've added a change which combines our approaches, what do you think? It should leave alone users who don't even use that example compose file, but warn if some of it is in use.

> This PR misses one instance of /etc/edumfa in [deploy/gunicorn/edumfaapp.py](https://github.com/eduMFA/eduMFA/blob/1352a7cc09945e531ccc6cd94006eac9a9a716ee/deploy/gunicorn/edumfaapp.py#L7).

`EDUMFA_CONFIGFILE` overrides those arguments, the config file is even the default argument for that method.  

> It is also a breaking change for anyone who had modified their logging config (see my comment there).

I've added the same check for the file and also a possibility to set its path from an environment variable.

> I would be very irritated if I had mounted my edumfa.cfg at /etc/edumfa and suddenly it disappears.

That would not happen due to the hashsum checks.

>  EDIT: Thinking about it, you probably can't delete a mounted file from inside the container in the first place...

At least in my testing it works fine since this is a volume, not a mount. Please do test and report back if you have different experiences!

aleyna72072: @tribut Do you still have interest in reviewing this? :slightly_smiling_face:

tribut: Sorry for not responding so long and thanks for the reminder. Hope to get to it tomorrow!

tribut (approved):
Thank you for taking the time to rework the patch. I understand that making the change as easy as possible for users is a concern and I agree with your approach.

LGTM!

deploy/docker/entrypoint.sh
tribut: What I meant regarding the disappearing file was a situation where the user had directly mounted the file. But it turns out, you cannot delete them in that case:

```bash
$ docker run --rm -it -v ./edumfa.cfg:/edumfa.cfg debian:stable-slim bash
root@1ddc7f4d4fbe:/# rm -f /edumfa.cfg 
rm: cannot remove '/edumfa.cfg': Device or resource busy
```

So it's not a concern.

aleyna72072:
@tribut Thank you for reviewing it, we'd love to see you around!

fritterhoff (approved):

PR #1095, opened 2026-04-14
docs(changelog): warn Ubuntu package users not to overwrite edumfa.cfg

aleyna72072:
Warn about apt wanting to replace `edumfa.cfg`. Maybe we should set the secrets separately from the config, as we do in the container image?  
This has caused at least confusion for two users.

### DIFF
--- a/READ_BEFORE_UPDATE.md
+++ b/READ_BEFORE_UPDATE.md
@@ -49,7 +49,7 @@ This release focuses on stability, security, and long-term maintainability. It c
 
 > [!CAUTION]
 >
-> Due to new fields in `edumfa.cfg`, updating your Ubuntu package will cause apt to ask you to replace it. Replacing that file will result in losing the secrets set in it (and you having to add those secrets back). If needed, add the new fields manually to your existing `edumfa.cfg`.
+> Due to new fields in `edumfa.cfg`, upgrading via Ubuntu packages prompts apt to replace that file. Replacing it can remove secrets from your current configuration, so keep your existing file and add only the new fields manually.
 >
 
 ### Added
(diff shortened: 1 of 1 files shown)


READ_BEFORE_UPDATE.md
```diff
@@ -47,6 +47,11 @@ This release focuses on stability, security, and long-term maintainability. It c
 > The 3.0.0 release of eduMFA is planned to remove multiple features. This will be announced as soon as there is a complete list. A work in progress list of removals can be found at: https://github.com/eduMFA/eduMFA/issues/875
 >
 
+> [!CAUTION]
+>
+> Due to new fields in `edumfa.cfg`, updating your Ubuntu package will cause apt to ask you to replace it. Replacing that file will result in losing the secrets set in it (and you having to add those secrets back). If needed, add the new fields manually to your existing `edumfa.cfg`.
```
Luc1412:
```suggestion
> Due to new fields in `edumfa.cfg`, upgrading via Ubuntu packages prompts apt to replace that file. Replacing it can remove secrets from your current configuration, so keep your existing file and add only the new fields manually.
```

Luc1412 (changes requested):
Looks good. I updated the PR title and improved the warning a bit to make it more concise.

READ_BEFORE_UPDATE.md
aleyna72072: I've applied your change with two small changes:
1. apt prompts the user, since apt itself can't be prompted if that makes sense
2. replaced can with will since it's unambiguous

Is that also fine with you? If you don't agree I would also be fine to just take your suggestion.

Luc1412 (approved):

aleyna72072:
related: https://github.com/eduMFA/eduMFA/issues/1124

PR #1097, opened 2026-04-17
change handling of entrypoint, use exec "$@" and CMD

belfhi:
It would be better to set the gunicorn command in the CMD of the dockerfile and only `exec "$@"` in the entrypoint script. This allows users to more easily customize the gunicorn arguments, for example if a different port should be used or ssl certificates should be provided. 

Also, it makes more sense to check if  EDUMFA_CONFIGFILE is not empty. Currently, if the environment variable is set but an empty string the logic fails.


aleyna72072 (approved):
Thanks, hope to see you around!

PR #1125, opened 2026-05-07
fix: radiusserver message authenticator

aleyna72072:
Fixes the Message-Authenticator for RADIUS requests. Tested with FreeRADIUS.  
Possible context:
- https://www.rfc-editor.org/rfc/rfc3579#page-16
- https://github.com/pyradius/pyrad/blob/2.5.4/pyrad/packet.py#L144-L149

### DIFF
--- a/edumfa/lib/radiusserver.py
+++ b/edumfa/lib/radiusserver.py
@@ -128,10 +128,12 @@ class RADIUSServer:
         try:
             response = srv.SendPacket(req)
             if config.enforce_ma:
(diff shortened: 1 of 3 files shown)


aleyna72072:
Drafted since it would be nice to have a unittest for this. The change itself should be fine.

aleyna72072:
@sklemer1 @fritterhoff Maybe you can at some point take a look at the comment in the unit test?

Luc1412 (approved):

PR #1136, opened 2026-05-13
fix: fudiscr error for unassigned passkeys

aleyna72072:
Unassigned Passkeys caused an error before, since "user" is already None. Also use the appropriate function for the orphan check. This would simply cause an error for the user logging in (fatal profile exception), so it shouldn't have any security implications.  
Tested orphaned user, unassigned token and normal Passkey login


aleyna72072:
@fritterhoff Pinging you for when you get back in case you have any remarks.

Luc1412 (approved):

fritterhoff (approved):

PR #1142, opened 2026-05-22
feat: Add multi-database test coverage and fix DB-specific query behavior

### LINKED ISSUE #1134: eduMFA 2.9.2 fails on Postgres database
### Top-level intent
eduMFA 2.9.0 ist running fine, after upgrade to 2.9.2 we get error 500 messages and database errors on postgres:

### Steps to reproduce

1. Ubuntu 24.04.4 LTS with Postgres 16 and eduMFA 2.9.0
2. upgrade eduMFA to 2.9.1 or 2.9.2
3. try to login to eduMFA

### Expected outcome
successful login

### Actual outcome
error 500 on web gui

What did actually happen?
2026-05-13 09:23:51.297 CEST [2605114] edumfa@edumfa ERROR:  FOR NO KEY UPDATE is not allowed with DISTINCT clause
2026-05-13 09:23:51.297 CEST [2605114] edumfa@edumfa STATEMENT:  SELECT DISTINCT token.id AS token_id, token.description AS token_description, token.serial AS token_serial, token.tokentype AS token_tokentype, token.user_pin AS token_user_pin, token.user_pin_iv AS token_user_pin_iv, token.so_pin AS token_so_pin, token.so_pin_iv AS token_so_pin_iv, token.pin_seed AS token_pin_seed, token.otplen AS token_otplen, token.pin_hash AS token_pin_hash, token.key_enc AS token_key_enc, token.key_iv AS token_key_iv, token.maxfail AS token_maxfail, token.active AS token_active, token.revoked AS token_revoked, token.locked AS token_locked, token.failcount AS token_failcount, token.count AS token_count, token.count_window AS token_count_window, token.sync_window AS token_sync_window, token.rollout_state AS token_rollout_state
        FROM token, realm, tokenrealm, tokenowner
        WHERE lower(realm.name) = 'xyz.de' AND tokenrealm.realm_id = realm.id AND tokenrealm.token_id = token.id AND tokenowner.realm_id = 3 AND tokenowner.token_id = token.id AND tokenowner.resolver = 'xxxxx' AND tokenowner.token_id = token.id AND tokenowner.user_id = 'xxxxxx' FOR NO KEY UPDATE

### Configuration

* **eduMFA version**: 2.9.2
* **Installation method**: from Ubuntu packages
* **Python version**: 3.12
* **Operating system**: Ubuntu 24.04.4 LTS
* **Webserver**: Apache2 
* **Token database**: PostgreSQL

fritterhoff:
Adds CI coverage for SQLite, PostgreSQL, and MariaDB-backed test runs, plus smoke tests using testcontainers for supported database backends.

Changes include:

- Add GitHub Actions jobs for PostgreSQL and MariaDB test matrices.
- Add testcontainers-based backend smoke tests.
- Add concurrent HOTP replay tests for database locking behavior.
- Adjust token realm filtering to avoid Postgres FOR UPDATE issues with joined tables.
- Fix MySQL audit date formatting percent escaping.
- Roll back failed ClientApplication inserts to keep the session usable.
- Relax backend-dependent test assumptions around collation, timestamps, and row counts.

Validation:

- Non-container tests exclude Docker-backed checks via the new dbcontainer marker.
- Database-specific checks now run separately against real Postgres/MariaDB backends.


.vscode/settings.json
Luc1412, line 1: This shouldn't be part of the PR

.github/workflows/test.yml
```diff
@@ -38,9 +38,97 @@ jobs:
...
+            files: ./coverage.xml
+
+  database-full-suite:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - database: postgres
+            image: postgres:16-alpine
```
aleyna72072: Why 16 in particular? psql is at 18 already, is this due to the AI not having data about newer versions or intentional because you want to ensure this versions works?

.github/workflows/test.yml
```diff
@@ -38,9 +38,97 @@ jobs:
...
+            port: 5432
+            test_database_url: postgresql+psycopg://edumfa:edumfa@127.0.0.1:5432/edumfa
+            health_cmd: pg_isready -U edumfa -d edumfa
+            extra_sync_args: ""
+            system_deps: ""
+          - database: postgres-psycopg2
+            image: postgres:16-alpine
```
aleyna72072: see above

.github/workflows/test.yml
```diff
@@ -38,9 +38,97 @@ jobs:
...
+            port: 5432
+            test_database_url: postgresql+psycopg2://edumfa:edumfa@127.0.0.1:5432/edumfa
+            health_cmd: pg_isready -U edumfa -d edumfa
+            extra_sync_args: "--extra postgres"
+            system_deps: "libpq-dev"
+          - database: mariadb
+            image: mariadb:11
```
aleyna72072: same question here

aleyna72072:
@fritterhoff This PR is more difficult to review, I'll ping you once I am done.

edumfa/lib/auditmodules/sqlaudit.py
```diff
@@ -92,7 +92,7 @@ def fn_to_isodate(element, compiler, **kw):
 def fn_to_isodate(element, compiler, **kw):
     # The four percent signs are necessary for two format substitutions
     return (
-        f"date_format({compiler.process(element.clauses, **kw)}, '%Y-%m-%d %H:%i:%s')"
+        f"date_format({compiler.process(element.clauses, **kw)}, '%%Y-%%m-%%d %%H:%%i:%%s')"
```
aleyna72072, line 95: What does this do?

.github/workflows/test.yml
aleyna72072: chat: this should probably be < 11.6.2 due to https://github.com/advisories/GHSA-qq2p-4282-cfc5

tests/test_lib_token.py
aleyna72072:
```suggestion
                # This tests if the parameter sortby is enforced. The result depends on the database collation, though.
                # return pagination
```

edumfa/lib/auditmodules/sqlaudit.py
```diff
@@ -92,7 +92,7 @@ def fn_to_isodate(element, compiler, **kw):
 def fn_to_isodate(element, compiler, **kw):
     # The four percent signs are necessary for two format substitutions
```
aleyna72072:
```suggestion
```

.vscode/settings.json
aleyna72072, line 1: Just to make sure: I don't use vscode, but I assume adding this is intentional as it's not part of https://github.com/github/gitignore/blob/main/Global/VisualStudioCode.gitignore ?

fritterhoff: It simplifies a lot ... could add the settings to the ignore but prop these settings are so generic

.github/workflows/test.yml
```diff
@@ -38,9 +38,92 @@ jobs:
...
+      matrix:
+        include:
+          - database: postgres-psycopg2
+            image: postgres:18-alpine
+            port: 5432
+            test_database_url: postgresql+psycopg2://edumfa:edumfa@127.0.0.1:5432/edumfa
+            health_cmd: pg_isready -U edumfa -d edumfa
+            extra_sync_args: "--extra postgres"
+            system_deps: "libpq-dev"
+          - database: mariadb
```
aleyna72072, line 61:
```suggestion
          # MariaDB version 11.5 should continue be tested for the foresee-able
          # future, as that version is vulnerable to
          # https://github.com/advisories/GHSA-qq2p-4282-cfc5 and testing with
          # it therefore makes sure the replayability fix keeps working.
          - database: mariadb
```
So when #1150 gets done we don't accidentally remove it.

tests/test_scripts.py
```diff
@@ -250,7 +250,7 @@ def test_05_edumfa_policy(self):
         assert result.exit_code == 0
         policies = Policy.query.all()
         assert len(policies) == 2
-        assert "Purged policy ui-totp with result 3" in result.output
+        assert "Purged policy ui-totp with result " in result.output
```
aleyna72072: On my machine it continues to work with the 3, is this on purpose?

aleyna72072:
Really nice to have these tests, thanks! Besides my two comments above I don't have any input.  
Let's wait for @Luc1412 wrt the CI and then do a 2.9.3 :rocket:  !

tests/test_scripts.py
fritterhoff: I assume you are stil running it against the sqlite? Irrc I had some issues in one of the CI runs.

.github/workflows/test.yml
fritterhoff: Well - in case we stop supporting it since it is EOL we should also remove it here...

tests/test_scripts.py
fritterhoff: image

Would guess the counter gets incremented in mariadb (and also postgres) over all tests in the pipline and sqlite resets it better - so sticking hard to the 3 is pretty strict. Could reset somehow the the counter before running the test, but from my point of view that is overkill and would blow up this PR even more.

fritterhoff:
> Really nice to have these tests, thanks! Besides my two comments above I don't have any input. Let's wait for @Luc1412 wrt the CI and then do a 2.9.3 🚀 !

@Luc1412 in theory we could run all three suits for both python versions (3.10 and 3.14) that would give the largest coverage, but I am not sure if it is worth...

.github/workflows/test.yml
aleyna72072: I assumed we'll still need it to make sure the replayability fix (for_update) keeps working. But looking back on https://github.com/aleyna72072/eduMFA/actions/runs/26649057879/job/78542358496?pr=1 since PostgreSQL in its default configuration is also vulnerable, we can probably just use that?

.github/workflows/test.yml
```diff
@@ -38,9 +38,92 @@ jobs:
...
+        with:
+          python-version: '3.14'
+      - name: Setup UV
+        uses: astral-sh/setup-uv@v8.1.0
+        with:
+          enable-cache: true
+      - name: Install dependencies
+        run: uv sync --group test --extra postgres
+      - name: Install postgres stuff
+        run: sudo apt-get install -y libpq-dev
```
Luc1412:
```suggestion
      - name: Install postgres dependencies
        run: sudo apt-get install -y libpq-dev
      - name: Install dependencies
        run: uv sync --group test --extra postgres
```

It's better to install system dependencies first

edumfa/lib/token.py
Luc1412, line 1: Wouldn't the removal of distinct cause issues with token counts when using owner filters?

tests/test_lib_token.py
```diff
@@ -1064,22 +1064,23 @@ def test_42_sort_tokens(self):
...
         self.assertTrue(tokens[0].get("serial") == "A8", tokens[0])
-        # SQLite does not sort like other DBs
-        if self.app.config["SQLALCHEMY_DATABASE_URI"].startswith("sqlite"):
-            self.assertEqual("hotptoken", tokens[-1].get("serial"), tokens[-1])
+        # MariaDB/MySQL uses case-insensitive collation (h < x), so 'X' sorts last.
+        # SQLite and PostgreSQL/Alpine use binary (C) collation (X=88 < h=104),
+        # so 'hotptoken' sorts last.
+        _db_uri = self.app.config["SQLALCHEMY_DATABASE_URI"]
+        if _db_uri.startswith("mysql"):
+            _last_asc = "X"
```
Luc1412, line 1071:
```suggestion
        if db.engine.dialect.name == "mysql":
```
This is the better way of checking dialect

tests/test_scripts.py
Luc1412: With this check becoming more loose, we maybe could use regex to match it for `> 0`

Luc1412 (commented):
I updated PR title and description. There are some improvements and considerations I found. Also not sure if some changes like `tests/test_api_validate_passkey.py` might be better in seperated commit/pr

edumfa/lib/token.py
fritterhoff: I don't think so. The amount should keep the same.

fritterhoff:
> I updated PR title and description. There are some improvements and considerations I found. Also not sure if some changes like `tests/test_api_validate_passkey.py` might be better in seperated commit/pr

Well. The test was failing several times. If you are fine with merging a PR that has unregullary failing tests I will revert the changes.

Luc1412 (approved):

aleyna72072 (approved):
LGTM

PR #1146, opened 2026-05-28
feat(webui): add panel for counting users with token

aleyna72072:
During a migration to eduMFA, users are often interested in the question: how many of my users have a token already? This PR provides one part of the answer: an interface to count how many users have a token of certain properties.  
image
It currently offers three possibilities of filtering, but (due to the way the query is built) should be easy to extend.  

This PR:
- adds a new documented API endpoint `/stats/current/users_with_token`, which will return the number of users having any kind of token
- adds a new function in `token.py` to count that
  + SimpleStats also uses that one now, which used a `subscriptions.py` function before (now removed)
  + this likely means that `subscriptions.py` can be easily removed now?
- unittests for both
- a new widget in the WebUI to check that endpoint, with the possibility to set parameters to narrow it down (e.g. only users with a certain tokentype)


aleyna72072:
@Luc1412 Especially interested in your input regarding design (and wording).

edumfa/static/templates/footer.html
```diff
@@ -40,6 +40,7 @@
 <script src="{{ instance }}/{{ "static/components/login/controllers/loginControllers.js" }}"></script>
 <script src="{{ instance }}/{{ "static/components/dashboard/controllers/dashboardControllers.js" }}"></script>
 <script src="{{ instance }}/{{ "static/components/dashboard/states/states.js" }}"></script>
+<script src="{{ instance }}/{{ "static/components/stats/factories/stats.js" }}"></script>
```
Luc1412: This file is missing

Luc1412 (changes requested):

edumfa/static/templates/footer.html
aleyna72072: Whoops, thanks for catching that! Added.

aleyna72072:
@Luc1412  ping, a user on `edumfa-users` asked exactly for this feature :smile: .

aleyna72072:
Introducing a new endpoint for this is admittedly not great, but putting this into a `/monitoring/(stats_key)` seemed worse.

edumfa/api/monitoring.py
```diff
@@ -119,3 +120,31 @@ def get_statistics_last(stats_key):
...
+    tokentype = getParam(param, "type", True)
+    realm = getParam(param, "realm", True)
+    active = getParam(param, "active", True)
+
+    if active:
+        active = active.lower() == "true"
+
+    return send_result(
+        count_users_with_token(realm=realm, active=active, tokentype=tokentype)
+    )
```
Luc1412: I think you missed a `g.audit_object.log({"success": True})`

edumfa/static/components/dashboard/views/dashboard.html
```diff
@@ -186,7 +186,48 @@
...
+                                    </select>
+                                </th>
+                                <th>
+                                    <label for="active" translate>Active</label>
+                                    <select class="form-control" ng-change="countUsersWithToken()"
+                                        ng-model="activenessForUsersWithToken">
+                                        <option><span translate>any</span></option>
+                                        <option><span translate>True</span></option>
+                                        <option><span translate>False</span></option>
+                                    </select>
```
Luc1412, line 211: Instead of active, we might better call it Status with values of `Any status`, `Active`, `Inactive`

edumfa/static/components/dashboard/views/dashboard.html
```diff
@@ -186,7 +186,48 @@
...
+                <div class="panel-heading" translate>Adoption</div>
+                <div class="panel-body">
+                    <table class="table">
+                        <tbody>
+                            <tr>
+                                <th>
+                                    <label for="realm" translate>Realm</label>
+                                    <select class="form-control" ng-change="countUsersWithToken()"
+                                        ng-model="realmForUsersWithToken">
+                                        <option><span translate>any</span></option>
```
Luc1412:
```suggestion
                                        <option value="any" translate>All Realms</option>
```

edumfa/static/components/dashboard/views/dashboard.html
```diff
@@ -186,7 +186,48 @@
...
+                                    <label for="active" translate>Active</label>
+                                    <select class="form-control" ng-change="countUsersWithToken()"
+                                        ng-model="activenessForUsersWithToken">
+                                        <option><span translate>any</span></option>
+                                        <option><span translate>True</span></option>
+                                        <option><span translate>False</span></option>
+                                    </select>
+                                </th>
+                                <th>
+                                    <label for="tokentype" translate>Tokentype</label>
```
Luc1412:
```suggestion
                                    <label for="tokentype" translate>Token Type</label>
```

edumfa/static/components/dashboard/views/dashboard.html
```diff
@@ -186,7 +186,48 @@
...
+                                    <select class="form-control" ng-change="countUsersWithToken()"
+                                        ng-model="tokentypeForUsersWithToken">
+                                        <option><span translate>any</span></option>
```
Luc1412:
```suggestion
                                        <option value="any" translate>Any type</option>
```

edumfa/static/components/dashboard/views/dashboard.html
```diff
@@ -186,7 +186,48 @@
...
+                                        <option ng-repeat="tokentype in tokentypes">{{ tokentype }}</option>
+                                    </select>
+                                </th>
+                            </tr>
+                            <tr>
+                                <td translate>Users having such a token:</td>
```
Luc1412:
```suggestion
                                <td translate>Users with matching token:</td>
```

Luc1412 (changes requested):
The implementation looks good. Maybe you can extend the PR description with more details e.g. what was changed in the backend.

When it comes to the implementation itself I got following annotions:
- You probably missed a audit log call?
- I improved the wording so it's more appealing to non-technical users
- There is currently no real loading state. Maybe this could become an issue on larger instances?

edumfa/static/components/dashboard/views/dashboard.html
aleyna72072: grafik
Changed it to Type, what do you think?

aleyna72072:
grafik

aleyna72072:
> There is currently no real loading state. Maybe this could become an issue on larger instances?

[Bildschirmaufnahme_20260701_181803.webm](https://github.com/user-attachments/assets/e29b0652-5eb2-462d-b6fb-6a87bc6b5229)

Maybe like this? Very simple, but maybe good enough?

edumfa/api/monitoring.py
aleyna72072: addded

edumfa/static/components/dashboard/views/dashboard.html
aleyna72072: added with disabled instead of inactive (consistency of capitalization and word)

Luc1412:
> > There is currently no real loading state. Maybe this could become an issue on larger instances?
> 
>  Bildschirmaufnahme_20260701_181803.webm 
> Maybe like this? Very simple, but maybe good enough?

Yeah that should be fine and helps the user to understand that it's working. Did you also handle error state properly? (resetting)

aleyna72072:
Missing:
- network error handling
- network timeout handling
- older requests shouldnt overwrite newer requests

aleyna72072:
@Luc1412 done

aleyna72072:
Tested it in a VM with roughly 41.000 users (SQLResolver on the same MariaDB on localhost) who each have a HOTP token:
[video](https://github.com/user-attachments/assets/8990a827-6261-4594-9b05-a1bf9c192099)  
FYI: Loading the dashboard on main already increases CPU load to 80/90% shortly.

With SQLite on the host, it takes a little bit more than double that time, but the dashboard does not really work with SQLite with default settings and that amount of users anyway.

aleyna72072:
@Luc1412 ping?

Luc1412:
LGTM, only translations need to be checked. Some pulled (probably wrong) fuzzy transaltions

Issue #1149, opened 2026-05-29
make sure migrations are tested


aleyna72072:
Table already gets created in https://github.com/eduMFA/eduMFA/blob/v2.9.2/tests/base.py#L66.
Make sure that upgrading them via the migrations are tested, too.  
Downgrading would be nice to have.

aleyna72072:
The tables would likely need to be pre-filled with data to catch migration errors.

PR #1151, opened 2026-05-30
feat: add a simulation of migration actions using github ci

fritterhoff:
More or less built on top of the #1142 and an attempt to provide a solution for #1149 

Right now only only python & only one mariadb/postgres version. More a proof of concept.


fritterhoff:
@aleyna72072 same as for the other PR ... exact versions must be discussed. Right now I stick on purpose on an old(er) python & mariadb version. If you really want to cover as much as possible we have to extend this to a `n x m` matrix.

.github/workflows/test.yml
```diff
@@ -43,4 +44,134 @@ jobs:
...
+          python -c 'import importlib.metadata as metadata; print("Release version:", metadata.version("edumfa"))'
+
+      - name: Create schema with latest release
+        run: |
+          set -euo pipefail
+
+          source "$VENV_DIR/bin/activate"
+          echo "Creating schema with released eduMFA"
+          edumfa-manage -q create_tables
+          edumfa-schema-upgrade "$VENV_DIR/lib/edumfa/migrations/" --skipstamp
```
aleyna72072, line 157: Did you test if this step actually fails if there is an exception during the migration?

aleyna72072: EDIT: Moved to #1172 to keep this focused. But the question above still applies for migrations not catching exceptions.

fritterhoff: s. https://github.com/eduMFA/eduMFA/actions/runs/27466712800/job/81190191335?pr=1151#step:11:35 -> CI fails :)

PR #1155, opened 2026-06-03
fix(image): fix edumfa-manage not finding config file

aleyna72072:
Currently, `edumfa-manage` works in the entrypoint and in user scripts because it exports `EDUMFA_CONFIGFILE`.  
It does not work when executing a shell in the container, as the default config-searching goes to `/etc/edumfa/edumfa.cfg`, and if it is not found, just [ignores](https://github.com/eduMFA/eduMFA/blob/v2.9.2/edumfa/app.py#L140) it. The default config file instead is at `/opt/edumfa/edumfa_config.py`.   
This fixes `edumfa-manage` in container shells by moving the config file search into edumfa itself in cases of containers (gated by an environment variable).

Although there are no problems reported, I assume this silently broke cronjobs which are supposed to rotate the audit log. Adding it to the v2.9.3 milestone because of this.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -40,6 +40,12 @@ RUN mkdir -p /opt/edumfa/user-scripts
 RUN python_package_path="$(python3 -c 'import site; x=site.getsitepackages(); assert len(x) == 1; print(x[0])')" && \
     ln -s "${python_package_path}/edumfa/" "/opt/edumfa/edumfa-package"
 
+# In order to enable edumfa-manage to automatically detect the correct config
+# file in a container image mark the images with an env variable.
+# This is checked in /edumfa/app.py
+# This is a workaround until 3.0.0.
+ENV __EDUMFA_RUNNING_IN_CONTAINER=1
+
 EXPOSE 8000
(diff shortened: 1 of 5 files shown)


fritterhoff:
~~Wondering how you executed the session in the container? Would it work to set the ENV?~~

Nvm. Forgot the tweak in the entrypoint.

aleyna72072:
> ~~Wondering how you executed the session in the container? Would it work to set the ENV? ~~

e.g. `docker exec docker-example-edumfa-1 edumfa-manage audit rotate`

> Nvm. Forgot the tweak in the entrypoint.

Yeah :/. Not the cleanest solution, but at least it's temporary.

aleyna72072:
@fritterhoff I removed the conf check, as I was not able to reproduce the crashes being delayed.

Luc1412 (approved):

fritterhoff (approved):
As discussed. Would prefer to rework/simplify the config parsing here in the future. Right now it is a bit confusing with 2x `silent=True`

PR #1159, opened 2026-06-08
fix(migrations): duplicate constraint on jwt blocklist nonce

aleyna72072:
The primary key constraint already enforces the nonce being unique. This does not cause problems on PostreSQL or MariaDB (which are **soon** the only supported DBs), but on Oracle DB.  
As in my testing this change does not make a difference on MariaDB or PostgreSQL, fix it for the users (which soon need to migrate anyway).


fritterhoff:
Does this require an migration?

aleyna72072:
Closing since while in theory the resulting behaviour is the same, the unique constraint actually results in the table on MariaDB:
```
PRIMARY KEY (`nonce`),
UNIQUE KEY `nonce` (`nonce`),
```

PR #1163, opened 2026-06-08
chore(docs): declare supported db systems

### LINKED ISSUE #1150: declare supported DBs
eduMFA (in my opinion) unofficially already has supported databases: those which the team can test (and are part of #1142). PRs which target other DBs can't be tested and are therefore have to lie dormant. This creates ambiguity and can lead to user frustration.  
This needs to be fixed by declaring the supported databases (versions pending) in the documentation:

- MariaDB
- PostgreSQL
- NOT SQLite, but it will continue to work due to it being used for development
  + there is no easy way to fix https://github.com/advisories/GHSA-qq2p-4282-cfc5 with SQLite 

The versions then need to be added to the tests in #1142.

aleyna72072:
In effect, eduMFA already has supported database systems: those we can test and therefore merge fixes for. This currently is not communicated effectively: PRs like #734 seem proper and ready to merge, but (due to missing development systems) are left to sit.  
Also looking at the recent security fixes in v2.9.2, the fine-grained behaviour of the database systems are relevant to ensure a secure system. Due to not having unlimited resources, this means narrowing down the scope.  

~~This PR is a **work in progess** to communicate the database systems which are supported.~~ Its core purpose is communicating what is already the case. The version definition goes a little beyond that, however is relevant in retrospect to the replayability issues.  

NOTE: This does not affect the SQL resolver! This is just about the database eduMFA uses for its own data.

TODO

- [x] decide whether this is the policy going forward
- [x] update docker.html and docker-example to use a supported db version
- [x] add the versions to the new DB test matrix
- [x] https://github.com/eduMFA/eduMFA/pull/1183/changes mysql faq; container; database.html
- [x] fix merge conflict
- [x] wait for #1256 to get merged

resolves #1150

### DIFF
--- a/doc/faq/mysqldb.rst
+++ b/doc/faq/mysqldb.rst
@@ -4,9 +4,10 @@ Database connect string
 -----------------------
 
 Due to its use of a database abstraction layer, eduMFA can work with several
-databases with the help of corresponding database drivers.
+databases with the help of corresponding database drivers. Please see
+:ref:`choosing_a_database` for supported databases.
 
-The database and corresponding diver are specified in the connect string
+The database and corresponding driver are specified in the connect string
 ``SQLALCHEMY_DATABASE_URI`` in :ref:`cfgfile`
 
 .. _mysqldb:
@@ -37,13 +38,3 @@ the eduMFA virtual environment with::
 The corresponding connect string looks like this:
 
 **connect string**: ``postgresql+psycopg2://<user>:<password>@<host>/<database>``
-
-
-Other databases
-~~~~~~~~~~~~~~~
-
-While we recommend MySQL as the backend database we regularly test MariaDB and
-PostgreSQL as well.
-
-Other databases like Oracle or MSSQL are working as well but not all
-functionality can be assured, so be aware that "Your mileage may vary".
--- a/doc/faq/mysqldb.rst
+++ b/doc/faq/mysqldb.rst
@@ -12,12 +12,15 @@ The database and corresponding driver are specified in the connect string
 
 .. _mysqldb:
 
-MySQL / MariaDB
-~~~~~~~~~~~~~~~
+MariaDB
+~~~~~~~
 
-While there are several database driver packages for MySQL, we recommend *PyMySQL*
-which is a pure Python package and does not require external libraries or a build
-environment on the server.
+While there are several database driver packages for MariaDB, we recommend
+*PyMySQL* which is a pure Python package and does not require external libraries
+or a build environment on the server.
+
+.. note:: While the name contains "MySQL", please note that MySQL (as opposed to
+         MariaDB) is not officially supported by eduMFA.
 
 *PyMySQL* is already installed in the virtual environment as a requirement for
 eduMFA.
--- a/doc/installation/database.rst
+++ b/doc/installation/database.rst
@@ -21,10 +21,6 @@ Notes:
 
 - None
 
-.. TODO: psql's default isolation level is Read Commited
-   (https://www.postgresql.org/docs/current/transaction-iso.html#XACT-READ-COMMITTED)
-   instead of Repeatable Read like in MariaDB. Is this an issue?
-
 MariaDB
 ^^^^^^^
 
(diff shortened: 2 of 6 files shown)


doc/installation/database.rst
```diff
@@ -0,0 +1,64 @@
...
+
+MariaDB is supported in version 11.8. Other versions may work too but are not
+tested.
+
+Notes:
+
+- Please make sure that ``innodb_snapshot_isolation`` is enabled. This is the
+  default starting from version 11.6.2.
+- If you are using Galera, avoid using a multi-primary setup. Try to only write
+  to one node at a time.
```
j-hoff: What's the reason for this recommendation? We are happy with our multi-master setup.

doc/installation/database.rst
```diff
@@ -0,0 +1,64 @@
...
+
+- the default version of Debian stable
+- the default version of the latest Ubuntu LTS
+- the latest LTS/stable version of the database system itself
+
+Of course, there will be a delay from a release of a new version until it is
+officially listed as supported. Older versions will also not instantly be kicked
+out to allow for a transitory period.
+
+Avoid these databases
```
aleyna72072: TODO: move this to the top as a generic "other dbs at own risk"

doc/installation/database.rst
```diff
@@ -0,0 +1,64 @@
...
+
+.. TODO: should it specify which isolation level to configure?
+
+Versions
+^^^^^^^^
+As a rule of thumb, eduMFA will support the following versions:
+
+- the default version of Debian stable
+- the default version of the latest Ubuntu LTS
+- the latest LTS/stable version of the database system itself
```
aleyna72072, line 41: accepted

doc/installation/database.rst
aleyna72072: TODO: remove this, mark Galera as supported

aleyna72072:
This was accepted as a policy on 2026-06-12 with the above change requirements.

aleyna72072:
Still draft because:  
MariaDB's default is REAPEATABLE READ (https://mariadb.com/docs/server/reference/sql-statements/transactions/transactions-repeatable-read), while PostgreSQL is READ COMMITTED (https://www.postgresql.org/docs/current/transaction-iso.html#XACT-READ-COMMITTED).  
As far as I remember, the isolation level was relevant for the fixes in 2.9.2/.1.  
@fritterhoff @j-hoff To take advantage of you having worked on that recently: should a transaction level be required?  

I _assume_ the current defaults are fine because on the above psql page it states:

> When a transaction uses this isolation level, a SELECT query (without a FOR UPDATE/SHARE clause) sees only data committed before the query began

And the race_condition unittests passes. Just want to make sure you see it the same way.


EDIT: discussed 2026-06-26 as something which is fine but is kept in mind

doc/installation/pip.rst
Luc1412: This footnote can be removed.

doc/faq/mysqldb.rst
Luc1412: This section might be a bit misleading now, since MySQL isn't supported but MariaDB using the MySQL dialect.

Luc1412 (changes requested):

doc/installation/pip.rst
aleyna72072: done

doc/faq/mysqldb.rst
aleyna72072: added a note and removed MySQL where it is possible

Luc1412 (dismissed):

doc/installation/docker.rst
```diff
@@ -85,11 +85,15 @@ To avoid the verbosity of ``docker run`` above, you can use a compose file.
...
+You could also use an external database or a different database system. For
+supported database systems and their versions, please see
+:ref:`choosing_a_database`.
+
 .. code-block:: yaml
 
    services:
      mariadb:
-       image: docker.io/mariadb:lts-noble
+       image: docker.io/mariadb:11.8
```
fritterhoff: Out of curiosity - why 11 and not going for 12?

aleyna72072: I just tested with that version and assumed it does not matter. Want me to change it?

aleyna72072: Re-tested with 12.3 and just changed it.

Luc1412 (approved):

PR #1006, opened 2026-02-15
feat: make challenge cleanup & client app tracking optional

fritterhoff:
## Top level description

This pull request adds new configuration options to allow administrators to disable client tracking and the automatic challenge janitor, as well as updates the UI and backend logic to support these options. It also introduces a CLI command for manual challenge cleanup and includes a test to verify the client tracking disablement.

Configuration and Feature Toggles:

* Added support for `DisableClientTracking` and `DisableAutoChallengeJanitor` configuration options, allowing administrators to disable client IP/user agent tracking and automatic challenge cleanup respectively in `edumfa/lib/clientapplication.py` and `edumfa/lib/tokenclass.py`. 

CLI and Maintenance:

* Exposed a new `cleanupchallenges` command in the `edumfa-token-janitor` CLI tool to manually trigger challenge cleanup, supporting environments where automatic cleanup is disabled.

PR #1164, opened 2026-06-09
chore(docs): document disabling the challenge janitor 

### LINKED ISSUE #1145: document DisableAutoChallengeJanitor
Introduced in https://github.com/eduMFA/eduMFA/pull/1006 but there is an entry missing in https://edumfa.readthedocs.io/en/v2.9.2/configuration/system_config.html.  
It should be noted this also affects the Passkey janitor: https://github.com/eduMFA/eduMFA/pull/1130/changes#diff-4682a5a779048b99dd96d778e4c8ceb40d7f6e992a09c35306a16f0601555eceR1645).

aleyna72072:
Updates the system configuration image and adds documentation for the two options introduced in #1006.  
Also fixes the wording, since since #1130 this also has an effect on the JWT blocklist.

fixes #1145


doc/configuration/system_config.rst
```diff
@@ -137,6 +137,31 @@ provides a wrong OTP value. AutoResync works like this:
 .. index:: usercache
 .. _user_cache_timeout:
```
Luc1412: With your changes, the anchor and index is now mispalced.

Luc1412 (changes requested):
LGTM just small issue with anchor.

doc/configuration/system_config.rst
aleyna72072: good catch, fixed!

Luc1412 (approved):

PR #1165, opened 2026-06-10
feat(webui): offer select_manually as a node for periodictasks

aleyna72072:
Periodic tasks have to specify a node to run on. This list is taken from two sources: the current `EDUMFA_NODE` and `EDUMFA_NODES`.  
In the Helm chart, cronjobs don't run on a node, a new container gets spawned to run the task. A workaround right now would be to specify some kind of placeholder in `EDUMFA_NODES`, but this does not seem like the correct way to approach this as it would not actually be an eduMFA node.  

Instead, this patch does it the other way around: it offers a "synthetic" option which then can be used with a manual override. It might also ease configuration outside of Helm for certain requirements.

### DIFF
--- a/doc/periodictask/index.rst
+++ b/doc/periodictask/index.rst
@@ -57,10 +57,11 @@ Every periodic task has the following attributes:
 	propagating the database changes to the other node). The name of the local node
 	as well as the names of remote nodes are configured in :ref:`cfgfile`.
 
-	Additionally, there is an option "select_manually". This option is
-	useful in case node names are unpredictable (like when used in a
-	Kubernetes deployment). In that case ``--node select_manually`` must be
-	passed to ``edumfa-cron``.
+	Additionally, there is a special option "select_manually" in the list of
+	nodes. This option is useful in case node names are unpredictable (like
+	when used in a Kubernetes deployment). If "select_manually" is ticked,
+	the task will run where ``edumfa-cron run_scheduled`` is executed with
+	the ``--node select_manually`` parameter.
 
 **taskmodule**
 	The task module determines the actual activity of the task. eduMFA comes
--- a/doc/periodictask/index.rst
+++ b/doc/periodictask/index.rst
@@ -57,6 +57,11 @@ Every periodic task has the following attributes:
 	propagating the database changes to the other node). The name of the local node
 	as well as the names of remote nodes are configured in :ref:`cfgfile`.
 
+	Additionally, there is an option "select_manually". This option is
+	useful in case node names are unpredictable (like when used in a
+	Kubernetes deployment). In that case ``--node select_manually`` must be
+	passed to ``edumfa-cron``.
+
 **taskmodule**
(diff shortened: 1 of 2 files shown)


Luc1412 (approved):

aleyna72072:
Requesting Jonas' review too, as he had questions.

j-hoff:
I just don't get how users of the UI are supposed to understand what the implication of ticking "select_manually" in the drop down is.
Where does the mnual selection of the node happen?

A second problem is, if you have for example three nodes in EDUMFA_NODES  (`mfa01-dev, mfa02-dev, mfa03-dev`) and hit "Select all" it would then select the four nodes: `mfa01-dev, mfa02-dev, mfa03-dev, select_manually"
Screenshot From 2026-06-11 10-45-33

aleyna72072:
> I just don't get how users of the UI are supposed to understand what the implication of ticking "select_manually" in the drop down is. Where does the mnual selection of the node happen?

When manually selected with `--node select_manually` in edumfa-cron as described in the docs part of this PR.

> A second problem is, if you have for example three nodes in EDUMFA_NODES (`mfa01-dev, mfa02-dev, mfa03-dev`) and hit "Select all" it would then select the four nodes: `mfa01-dev, mfa02-dev, mfa03-dev, select_manually"

That would not be a problem imo - this does not do anything except when the user manually selects "select_manually".

But happy to take any suggestions on how to improve this PR!

aleyna72072:
For future travellers: a nicer solution would be some kind of "mark system" in addition/alternative to the node constraint for the task scheduling.    
At the moment, this change seems to be too invasive (migration, UI and code changes in edumfa and edumfa-cron) for "just trying to make edumfa-cron work in a Helm context".

doc/periodictask/index.rst
```diff
@@ -57,6 +57,11 @@ Every periodic task has the following attributes:
 	propagating the database changes to the other node). The name of the local node
 	as well as the names of remote nodes are configured in :ref:`cfgfile`.
 
+	Additionally, there is an option "select_manually". This option is
+	useful in case node names are unpredictable (like when used in a
+	Kubernetes deployment). In that case ``--node select_manually`` must be
+	passed to ``edumfa-cron``.
```
j-hoff:
```suggestion
Additionally, there is a special node "select_manually" in the list. This option is 
useful in case node names are unpredictable (like when used in a Kubernetes 
deployment). If "select_manually" is ticked, the task will run on the nodes 
where ``edumfa-cron run_scheduled`` is run with the ``--node select_manually`` 
parameter.
```

j-hoff (changes requested):

doc/periodictask/index.rst
j-hoff: somehow the code suggestion is messed up, sorry, 
Supposed to replace L60-L63 of course.

aleyna72072: applied with slight changes

j-hoff (approved):
lgtm now

PR #1168, opened 2026-06-10
chore: sort container keys

aleyna72072:
Sorts some keys so it's nicer to look at in the documentation as well as the example docker file.  
In the documentation, it's only sorted inside the categories (MariaDB, required and optional env vars).


aleyna72072:
@Luc1412 ready to review now!

Luc1412 (approved):

PR #1173, opened 2026-06-12
chore(ci): test supported db versions

aleyna72072:
Adds the versions which were determined in #1163 to be supported DBs to the test matrix.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,19 +51,40 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - database: postgres-psycopg2
+          - database: postgres-17-psycopg2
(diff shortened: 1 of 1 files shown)


fritterhoff (approved):
Hope we don't get much more versions; Otherwise we prop should consider to rework it again.

aleyna72072:
> Hope we don't get much more versions; Otherwise we prop should consider to rework it again.

It's likely we get one more for psql and MariaDB for the transitory period between Debian and Ubuntu releases. Are we running into some kind of limit?

fritterhoff:
No. No real limit. But we scale by duplicating the lines ... maybe we can split the stuff or something like that

PR #1174, opened 2026-06-12
feat(helm): add helm chart

aleyna72072:
A Helm chart for eduMFA. It works via a deployment and a Helm job which runs on install/upgrade. This job updates the database schema, so it's not done by multiple containers at the same time.  
The deployment also uses "Recreate" as a strategy to make sure two eduMFA versions are not running at the same time.  
This PR is a draft to collect feedback.  

To try out this chart, see the "helm.rst" part of this PR. It requires using a pre-release container image, as the entry point had to be modified. That pre-release image can be found in the `values.yaml`.

TODO:

- [x] add #1175 as configurable readiness-/livenessprobe
- [x] CI: build and push chart
- [x] figure out how to link to values.yaml from documentation
- [x] add a link to supported dbs in `helm.rst`
- [x] :exclamation: This will move to its own repo once all parties approve this chart.


aleyna72072:
@belfhi @hajo-m @bit-project-de would love to get your feedback!

fritterhoff:
Just a few ideas from side:

- We could add `helm.sh/hook-delete-policy` to the jobs so hooks are only deleted on `hook-succeeded`
- We could add a `health-check` job so we can 'break' the helm deployment in case sth is going wrong (e.g. performing a curl) 
- Even though it is out of scope: We should consider adding a real `/health` or `/ready` endpoint...
- From my experience DB migrations are commonly done in the pre-upgrade step but that is a long discussion.
- Maybe we can split up loading the secrets a little bit. (At least the db secret - so you can refer to an existing secret there)

doc/installation/helm.rst
```diff
@@ -0,0 +1,71 @@
+.. _install_helm:
+
+Helm Installation
+-------------------
+
+.. index:: helm
+
+.. warning:: This chart is currently considered experimental. This means it does
+             not offer the semantic versioning guarantees: breaking changes can
+             happen in any release.
```
Luc1412: I think this comment regarding version guarantees is unnecessary. As long the major version is 0, you're allowed to publish breaking changes with every minor version.

aleyna72072: I want to make sure this is very clear to any interested party. I've changed the wording a bit, is it maybe better like this?

aleyna72072:
>     * We could add `helm.sh/hook-delete-policy` to the jobs so hooks are only deleted on `hook-succeeded`

I guess that makes sense? Normal users probably don't  care what a succesful migration did. I wonder if that could be a problem if used with automation (e.g. fluxcd or argocd). Wouldn't want the jobs to keep piling up.  
What do you think? If we go this route, maybe we should also add a TTL?

>     * We could add a `health-check` job so we can 'break' the helm deployment in case sth is going wrong (e.g. performing a curl)

This can probably just be done as part of the livenessProbe?

>     * Even though it is out of scope: We should consider adding a real `/health` or `/ready` endpoint...

Curious onlookers: see #1175 

>     * From my experience DB migrations are commonly done in the pre-upgrade step but that is a long discussion.

But this would cause the migration to be performed while the old eduMFA version is still running? Currently, the timeframe is much smaller and during it, the old deployment is in terminating stage. If I understand [gunicorn](https://gunicorn.org/signals/) correctly, it should not accept new connections during it, too.  
It would be really cool to be able to give some guarantees which would enable zero-downtime upgrades on patch releases via the Rolling Update strategy, but this seems a bit too ambitious at this stage. Also relying on operators to scale down before applying other updates seems like recipe for breakage (although I can't exactly say how this breakage in the DB would look like from the offhand).

>     * Maybe we can split up loading the secrets a little bit. (At least the db secret - so you can refer to an existing secret there)

I can see that, but I am wondering how the chart then knows where to find the data item? If it were just psql, it could pick `POSTGRES_PASSWORD`, but since two DBs are supported...

aleyna72072:
v0.1.1: fixed podannotations

belfhi:
I would probably pull a few configs out and set them as defaults in the values.yaml. For example in the spec:

```yaml
          {{- if .Values.healthChecks.enabled }}
          livenessProbe:
            {{- with .Values.healthChecks.livenessProbe }}
            {{- tpl . $ | nindent 12 }}
            {{- end }}
          readinessProbe:
            {{- with .Values.healthChecks.readinessProbe }}
            {{- tpl . $ | nindent 12 }}
            {{- end }}
          {{- end }}
```
and then set sensible defaults in values.yaml:
```yaml
healthChecks:
  enabled: true
  livenessProbe: |
    httpGet:
      path: /
      scheme: HTTP
      port: '{{ .Values.service.port }}'
  readinessProbe: |
    httpGet:
      path: /
      scheme: HTTP
      port: '{{ .Values.service.port }}'
```
this way you allow for customization (e.g. if someone wants to use https in the backend, too).
With the secrets I agree that splitting them up makes sense. Allow for setting the secretKey for the existingsecret. That wat users can create a single secret for all or use 
```yaml
            {{- if .Values.database.existingSecret -}}
           - name: DB_PASSWORD
              valueFrom:
                secretKeyRef:
                  name: {{ .Values.database.existingSecret }}
                  key: {{ .Values.database.existingSecretKey | default "password" }}
            {{- end }}
```
or something like that.

aleyna72072:
v0.1.2: db password secret is now separate

aleyna72072:
@belfhi Thanks! The configurable probes will have to wait for #1175, but I've added your other suggestion.

doc/installation/helm.rst
Luc1412: Looks better now. You could additionally specifically refer to minor version including breaking changes

aleyna72072:
v0.2.0:

- liveness- and readinessProbe are now configurable
- admin credentials can be set from values.yaml
- breaking change: changed the path/names for the db secret for uniformity

Please use the new `values.yaml`.

doc/installation/helm.rst
```diff
@@ -0,0 +1,70 @@
...
+- Move the key to your current directory.
+2. Create the audit key
+
+- Create the key: ``uv run -w edumfa --prerelease=allow edumfa-manage create_audit_keys``
+- Move the certificate and key to your current directory.
+3. Create a secret key for the JWTs: ``$ tr -dc A-Za-z0-9_ </dev/urandom | head -c24 > secret_key``
+4. Create a pepper: ``$ tr -dc A-Za-z0-9_ </dev/urandom | head -c24 > pepper``
+5. Create a Kubernetes secret from these files: ``$ kubectl create secret generic edumfa-secrets --from-file=. -n $YOUR_NAMESPACE``
+6. **Backup the files in this directory before deleting them!**  Especially the enckey is necessary to restore from a database backup as token secrets are encrypted.
+7. After **backing them up** you may delete the directory and files.
```
aleyna72072, line 58: TODO: show example secret

aleyna72072:
moved to:  https://github.com/eduMFA/helm-charts/tree/main/charts/edumfa#edumfa-chart  
Please note 0.3.x and 0.4.x were skipped. Check the changelog before updating to 0.5.0.

aleyna72072:
The installation instructions are maintained in the above repo instead.  
https://github.com/eduMFA/eduMFA/pull/1266

PR #1175, opened 2026-06-13
feat: add an (unauthenticated) health check endpoint

fritterhoff:
Adds unauthenticated Flask health endpoints for orchestration probes.

`/health/ready` now verify database connectivity, `/health/live` provides a lightweight liveness endpoint. Includes focused tests for live, DB-only readiness.

---
<a href="https://app.blacksmith.sh/eduMFA/codesmith/eduMFA/pr/1175"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1783926546&installation_id=137412999&pr_number=1175&repository=eduMFA%2FeduMFA&return_to=https%3A%2F%2Fgithub.com%2FeduMFA%2FeduMFA%2Fpull%2F1175&signature=8a56217872807b6f2b4af341f6375078ff0c12ceab663b104a6a17db164fca13"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/edumfa/api/health.py
+++ b/edumfa/api/health.py
@@ -27,8 +27,8 @@ from alembic.config import Config
 from alembic.runtime.migration import MigrationContext
 from alembic.script import ScriptDirectory
 from flask import Blueprint, Response, jsonify, request
-from sqlalchemy import text
 
+from edumfa.lib.sqlutils import is_db_available
 from edumfa.models import db
 
 health_blueprint = Blueprint("health_blueprint", __name__)
@@ -66,14 +66,6 @@ def _get_database_schema_head(connection) -> str | None:
     return context.get_current_revision()
 
 
-def _check_database(connection) -> dict[str, str]:
-    """
-    Check whether a simple query can be executed against the database.
-    """
-    connection.execute(text("SELECT 1"))
-    return {"status": "ok"}
-
-
 def _check_schema(connection) -> dict[str, str | None]:
     """
     Check whether the database schema revision matches the application migrations.
@@ -91,24 +83,30 @@ def _check_schema(connection) -> dict[str, str | None]:
     return result
 
 
-def _health_response(check_schema: bool = False) -> tuple[object, int]:
+def _health_response(check_schema: bool = False) -> tuple[Response, int]:
     """
     Build the readiness response and HTTP status code from the configured checks.
     """
     status_code = 200
     checks: dict[str, dict[str, str | None]] = {}
+    database_available = is_db_available(db.engine)
+
+    if database_available:
+        checks["database"] = {"status": "ok"}
+    else:
+        checks["database"] = {
+            "status": "error",
+            "message": "Database connection check failed",
+        }
+        return jsonify({"status": "error", "checks": checks}), 503
 
     try:
-        with db.engine.connect() as connection:
-            checks["database"] = _check_database(connection)
-            if check_schema:
+        if check_schema:
+            with db.engine.connect() as connection:
                 checks["schema"] = _check_schema(connection)
     except Exception as exx:
         status_code = 503
-        if "database" not in checks:
-            checks["database"] = {"status": "error", "message": str(exx)}
-        elif check_schema:
-            checks["schema"] = {"status": "error", "message": str(exx)}
+        checks["schema"] = {"status": "error", "message": str(exx)}
 
     if any(check["status"] != "ok" for check in checks.values()):
         status_code = 503
@@ -134,7 +132,7 @@ def live() -> Response:
 
 
 @health_blueprint.route("/ready", methods=["GET"])
-def ready() -> Response:
+def ready() -> tuple[Response, int]:
     """
     Return readiness based on database availability and optional schema state.
     """
--- a/edumfa/api/health.py
+++ b/edumfa/api/health.py
@@ -1,0 +1,116 @@
+#
+# License:  AGPLv3
+# This file is part of eduMFA. eduMFA is a fork of privacyIDEA which was forked from LinOTP.
+# Copyright (c) 2026 eduMFA Project-Team
+#
+# This code is free software; you can redistribute it and/or
+# modify it under the terms of the GNU AFFERO GENERAL PUBLIC LICENSE
+# License as published by the Free Software Foundation; either
+# version 3 of the License, or any later version.
+#
+# This code is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+#
+# You should have received a copy of the GNU Affero General Public
+# License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Unauthenticated health endpoints for load balancers and orchestration probes like kubernetes, haproxy, nginx or docker.
+"""
+
+from pathlib import Path
+from sysconfig import get_paths
+
+from alembic.config import Config
+from alembic.runtime.migration import MigrationContext
+from alembic.script import ScriptDirectory
+from flask import Blueprint, jsonify, request
+from sqlalchemy import text
+
+from edumfa.models import db
+
+health_blueprint = Blueprint("health_blueprint", __name__)
+
+
+def _migration_directory():
+    candidates = [
+        Path(__file__).resolve().parents[2] / "migrations",
+        Path(get_paths()["data"]) / "lib" / "edumfa" / "migrations",
+    ]
+    for candidate in candidates:
+        if (candidate / "env.py").is_file() and (candidate / "versions").is_dir():
+            return candidate
+    raise FileNotFoundError("Could not find eduMFA migration directory")
+
+
+def _get_required_schema_heads():
+    migrations = _migration_directory()
+    config = Config(str(migrations / "alembic.ini"))
+    config.set_main_option("script_location", str(migrations))
+    return sorted(ScriptDirectory.from_config(config).get_heads())
+
+
+def _get_database_schema_heads(connection):
+    context = MigrationContext.configure(connection)
+    return sorted(context.get_current_heads())
+
+
+def _check_database(connection):
+    connection.execute(text("SELECT 1"))
+    return {"status": "ok"}
+
+
+def _check_schema(connection):
+    required_heads = _get_required_schema_heads()
+    current_heads = _get_database_schema_heads(connection)
+    result = {
+        "status": "ok",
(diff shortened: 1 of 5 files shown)


fritterhoff:
@aleyna72072 the schema check is pretty hacky ... could remove that fragment totally but maybe it makes sense?

fritterhoff:
Could be reused in #1174

edumfa/api/health.py
```diff
@@ -0,0 +1,116 @@
...
+        if "database" not in checks:
+            checks["database"] = {"status": "error", "message": str(exx)}
+        elif check_schema:
+            checks["schema"] = {"status": "error", "message": str(exx)}
+
+    if any(check["status"] != "ok" for check in checks.values()):
+        status_code = 503
+
+    status = "ok" if status_code == 200 else "error"
+    return jsonify({"status": status, "checks": checks}), status_code
```
aleyna72072, line 115: This does not use the standard `send_result` method. Since that is the case everywhere else: is there a reason why this can't be used here?

edumfa/api/health.py
```diff
@@ -0,0 +1,116 @@
...
+from alembic.script import ScriptDirectory
+from flask import Blueprint, jsonify, request
+from sqlalchemy import text
+
+from edumfa.models import db
+
+health_blueprint = Blueprint("health_blueprint", __name__)
+
+
+def _migration_directory():
```
aleyna72072: How safe is this? Would love to see it in edumfa-schema-update if it is.

edumfa/api/health.py
fritterhoff: Wanted to keep it as basic as possible. The `send_result` is adding a bunch of additional information. Would guess that it prop. should work ...

edumfa/api/health.py
fritterhoff: As mentioned that is more or less a hacky attempt based on some StackOverflow and GitHub (and a mini portion AI) inspiration ... 

At least in my attempts that was working pretty well. The `schema-update` would prop. need some more quirks since the binary is installed to `/opt/edumfa` in case of the ubuntu image.

edumfa/api/health.py
aleyna72072: Yes, and especially `prepare_result` sucks right now.. but that's something we need to fix anyway someday.  
see #1137

edumfa/api/health.py
```diff
@@ -0,0 +1,116 @@
...
+    config.set_main_option("script_location", str(migrations))
+    return sorted(ScriptDirectory.from_config(config).get_heads())
+
+
+def _get_database_schema_heads(connection):
+    context = MigrationContext.configure(connection)
+    return sorted(context.get_current_heads())
+
+
+def _check_database(connection):
```
aleyna72072: https://github.com/eduMFA/eduMFA/blob/v2.9.3/edumfa/commands/manage/core.py#L249  
Would be nice to have this as a shared method instead.

fritterhoff: Mhm. The existing method returns with sys.exit(1) - that would prop kill the process? Could refactor it a bit ...

edumfa/api/health.py
```diff
@@ -0,0 +1,116 @@
...
+        if (candidate / "env.py").is_file() and (candidate / "versions").is_dir():
+            return candidate
+    raise FileNotFoundError("Could not find eduMFA migration directory")
+
+
+def _get_required_schema_heads():
+    migrations = _migration_directory()
+    config = Config(str(migrations / "alembic.ini"))
+    config.set_main_option("script_location", str(migrations))
+    return sorted(ScriptDirectory.from_config(config).get_heads())
```
aleyna72072: This should imo be [ScriptDirectory.get_current_head()](https://alembic.sqlalchemy.org/en/latest/api/script.html#alembic.script.ScriptDirectory.get_current_head) instead.  
https://alembic.sqlalchemy.org/en/latest/api/script.html#alembic.script.ScriptDirectory.get_heads

edumfa/api/health.py
```diff
@@ -0,0 +1,116 @@
...
+
+def _check_schema(connection):
+    required_heads = _get_required_schema_heads()
+    current_heads = _get_database_schema_heads(connection)
+    result = {
+        "status": "ok",
+        "current": current_heads,
+        "required": required_heads,
+    }
+    if current_heads != required_heads:
```
aleyna72072: see above, if it is more than one head this check does not seem to make sense?

edumfa/api/health.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+@health_blueprint.route("/live", methods=["GET"])
+def live():
+    return jsonify({"status": "ok"})
+
+
+@health_blueprint.route("", methods=["GET"])
+@health_blueprint.route("/", methods=["GET"])
+@health_blueprint.route("/ready", methods=["GET"])
```
aleyna72072: I think I would prefer if this would just be `/ready`? That way we still have wiggle room to put something in `/` without a breaking change.

aleyna72072:
And to get ahead of Lucas: type hints and docstrings are missing :smile: .

aleyna72072:
added to milestone 2.10.0 for the helm chart release

aleyna72072:
> @aleyna72072 the schema check is pretty hacky ... could remove that fragment totally but maybe it makes sense?

I mean it kinda makes sense? An eduMFA version should always be run with the database schema it's intended to run with, everything else is probably broken.  
Maybe let's even make it non-optional?

EDIT: sorry for closing, slipped

fritterhoff:
Right now I have not found a better solution for this check ... it seems to work in the most deployments but in case not I would leave it optional to gather some insights.

fritterhoff:
Ping?

aleyna72072:
I assumed this PR is not finished yet because of the draft state?

fritterhoff:
Regarding the discussion on the version/stamp information. Could remove it...

edumfa/api/health.py
```diff
@@ -0,0 +1,139 @@
...
+                checks["schema"] = _check_schema(connection)
+    except Exception as exx:
+        status_code = 503
+        checks["schema"] = {"status": "error", "message": str(exx)}
+
+    if any(check["status"] != "ok" for check in checks.values()):
+        status_code = 503
+
+    status = "ok" if status_code == 200 else "error"
+    return jsonify({"status": status, "checks": checks}), status_code
```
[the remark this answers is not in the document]
aleyna72072: This concern is in-line with what was discussed some Friday ago.

edumfa/api/health.py
aleyna72072: ping?

aleyna72072:
> Regarding the discussion on the version/stamp information. Could remove it...

Maybe do that for now to keep things simple and get this merged? Then we can revisit it at a later point

aleyna72072 (changes requested):
see above

edumfa/api/health.py
fritterhoff: Would keep it as simple as possible. as mentioned `send_result` has some overhead and sideeffects.

aleyna72072: In general I agree with keeping things simple, but that method is used everywhere eduMFA returns something via the API (which is great consistency). Is there a reason why the healthcheck should be special?  
Improving `send_results` is a separate topic, of course.

tests/test_api_health.py
```diff
@@ -0,0 +1,37 @@
...
+    def test_00_live_health_is_unauthenticated(self):
+        with self.app.test_request_context("/health/live", method="GET"):
+            res = self.app.full_dispatch_request()
+            self.assertEqual(200, res.status_code, res.data)
+            self.assertEqual({"status": "ok"}, res.json)
+
+    def test_01_ready_health_checks_database(self):
+        with self.app.test_request_context("/health/ready", method="GET"):
+            res = self.app.full_dispatch_request()
+            self.assertEqual(200, res.status_code, res.data)
```
aleyna72072, line 38: Is it (easily) possible to also test with a unavailable database to ensure the health check correctly fails? This would also implicitly test the new method (`is_db_available`) I think.

aleyna72072 (approved):
Looks great, but the two questions above would be nice to have an answer for.

aleyna72072 (dismissed):
It would be nice to decide on the two issues first.

edumfa/api/health.py
aleyna72072: @Luc1412 Looking at this thread, we've done a circle with regards to arguments. To avoid doing another round and also conflict: care to weigh in?  
If you don't have an opinion on this, @fritterhoff feel free to merge this with a bypass. It probably isn't that deep after all.

tests/test_api_health.py
Luc1412, line 1: The tests only returns successful paths. Could you add a 503 path with mocking

edumfa/api/health.py
```diff
@@ -0,0 +1,48 @@
...
+# GNU AFFERO GENERAL PUBLIC LICENSE for more details.
+#
+# You should have received a copy of the GNU Affero General Public
+# License along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+"""
+Unauthenticated health endpoints for load balancers and orchestration probes like kubernetes, haproxy, nginx or docker.
+"""
+
+from flask import Blueprint, Response, jsonify, request
```
Luc1412: request is an unused import

Luc1412 (changes requested):
The implementation looks good. Found just some minor mistakes.

Additionally the PR description is outdated. It mentions an nonexistent endpoint `/health` and also nonexistent query `?schema=true`.

Regarding the `jsonify` vs. `send_result`,  I'd prefer with `jsonify`. The endpoint is intended to be consumed by k8s or docker. A minimal stable response payload is the better choice here. The json RPC data is irrelevant to those systems. Additionally it prevents weird results like this this:
```json
{
  "result": {
    "status": true,
    "value": {
      "status": "error"
    }
  }
}
```

tests/test_api_health.py
aleyna72072: resolving in favour of the same request by Lucas below

Luc1412 (approved):

PR #1183, opened 2026-06-16
feat: add hints to use read_comitted for maria db

### LINKED ISSUE #1182: Unreliable `pytest` for MariaDB 12.3.
Due to unknown reason the test is a bit flacky for MariaDB 12.3

s. https://github.com/eduMFA/eduMFA/actions/runs/27617464043/job/81656938709#step:8:4652

```
------------------------------ Captured log call -------------------------------
INFO     edumfa.lib.crypto:crypto.py:835 initializing HSM class: <class 'edumfa.lib.security.default.DefaultSecurityModule'>
INFO     edumfa.lib.crypto:crypto.py:256 Initialized HSM object {'obj': <edumfa.lib.security.default.DefaultSecurityModule object at 0x7f38ed4c7cd0>}
WARNING  edumfa.lib.config:config.py:374 Request-local store already contains config object, even though it should not
WARNING  edumfa.models:models.py:2749 Unable to update ClientApplication entry: (pymysql.err.OperationalError) (1020, "Record has changed since last read in table 'clientapplication'; try restarting transaction")
[SQL: UPDATE clientapplication SET lastseen=%(lastseen)s WHERE clientapplication.id = %(id_1)s]
[parameters: {'lastseen': datetime.datetime(2026, 6, 16, 12, 37, 38, 452997), 'id_1': 1}]
(Background on this error at: https://sqlalche.me/e/20/e3q8)
WARNING  edumfa.models:models.py:2749 Unable to update ClientApplication entry: (pymysql.err.OperationalError) (1020, "Record has changed since last read in table 'clientapplication'; try restarting transaction")
[SQL: UPDATE clientapplication SET lastseen=%(lastseen)s WHERE clientapplication.id = %(id_1)s]
[parameters: {'lastseen': datetime.datetime(2026, 6, 16, 12, 37, 38, 455040), 'id_1': 1}]
(Background on this error at: https://sqlalche.me/e/20/e3q8)
WARNING  edumfa.models:models.py:2749 Unable to update ClientApplication entry: (pymysql.err.OperationalError) (1020, "Record has changed since last read in table 'clientapplication'; try restarting transaction")
[SQL: UPDATE clientapplication SET lastseen=%(lastseen)s WHERE clientapplication.id = %(id_1)s]
[parameters: {'lastseen': datetime.datetime(2026, 6, 16, 12, 37, 38, 465618), 'id_1': 1}]
(Background on this error at: https://sqlalche.me/e/20/e3q8)
INFO     edumfa.lib.tokens.hotptoken:hotptoken.py:509 Previous OTP used again. Serial RACEHOTP01 with counter 2.
ERROR    edumfa.api.before_after:before_after.py:432 Database error occurred: OperationalError('(pymysql.err.OperationalError) (1020, "Record has changed since last read in table \'token\'; try restarting transaction")')
INFO     edumfa.lib.tokens.hotptoken:hotptoken.py:509 Previous OTP used again. Serial RACEHOTP01 with counter 2.
INFO     edumfa.lib.tokens.hotptoken:hotptoken.py:509 Previous OTP used again. Serial RACEHOTP01 with counter 2.
=============================== warnings summary ===============================
```

fritterhoff:
This pull request introduces configuration and documentation updates to improve database reliability, especially for MySQL/MariaDB environments. The main focus is on explicitly setting the transaction isolation level to `READ COMMITTED` to prevent errors. Additionally, the CI workflow is updated to use these configurations during testing.

**Database reliability and configuration improvements:**

* Updated documentation in `doc/faq/mysqldb.rst` and `doc/installation/system/inifile.rst` to recommend and explain setting `SQLALCHEMY_ENGINE_OPTIONS = {"isolation_level": "READ COMMITTED"}` for MySQL/MariaDB, including a new section on transaction isolation levels and their impact. 
* Added guidance and configuration for `EDUMFA_AUDIT_SQL_OPTIONS = {}` when the audit database uses a different backend (e.g., SQLite) to avoid compatibility issues. 
* Updated the example and default configuration in `deploy/edumfa/edumfa.cfg` to include comments and options for setting the isolation level and audit SQL options. 

**Continuous integration and testing:**

* Modified `.github/workflows/test.yml` to ensure the correct `edumfa.cfg` is used for MariaDB jobs, and to pass the configuration file path as an environment variable during tests. 
* Updated `tests/testdata/edumfa-mariadb.cfg` to explicitly set `READ COMMITTED` isolation level and reset audit SQL options, keeping CI in sync with documentation and best practices.This pull request changes the default isolation level to `READ COMMITTED` for MariaDB/MySQL, addressing concurrency issues during token validation. You may opt out using the SQLAlchemy Engine options.

**Database transaction isolation configuration:**

* Added logic in `edumfa/app.py` to set the SQLAlchemy engine's `isolation_level` to `READ COMMITTED` for MariaDB/MySQL, preventing errors during concurrent token validation. It checks for existing settings and does not perform any change if nothing is required. 

---
<a href="https://app.blacksmith.sh/eduMFA/codesmith/eduMFA/pr/1183"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1784221935&installation_id=137412999&pr_number=1183&repository=eduMFA%2FeduMFA&return_to=https%3A%2F%2Fgithub.com%2FeduMFA%2FeduMFA%2Fpull%2F1183&signature=25f0622c2c498413e685f2ecccc104d8f7b9cdf86876fb6a8a53b350c5cff788"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


aleyna72072:
Why is this opt-in? I'd consider this a bug fix and also - imo - reducing the number of possible options in eduMFA would be good.

doc/installation/system/inifile.rst
```diff
@@ -86,6 +86,28 @@ slower but more robust and can be necessary in large redundant setups.
    row is otherwise not present". In this case setting ``EDUMFA_DB_SAFE_STORE``  to *True*
    might help.
 
+.. _mysql_isolation_level:
+
+If your database is MySQL or MariaDB, eduMFA automatically uses the
+``READ COMMITTED`` transaction isolation level instead of the MySQL/MariaDB
+default ``REPEATABLE READ``. With ``REPEATABLE READ`` a consistent read
```
aleyna72072: This is misleading I think, since operators might configure a different isolation level on PostgreSQL, too.  

Maybe instead:
> eduMFA automatically uses the ``READ COMMITTED`` transaction isolation level by default, even if you configure your database system differently. With ``REPEATABLE READ`` a consistent read [...]

fritterhoff: Right now we only replace the default level for MariaDB since that is the default level. I would keep the change as small as possible and only change the current default settings.

aleyna72072: Ah, didn't see that as it was further below. With the goal of having less diverse configurations (and as few special cases for our supported DBs), I fail to find reasons why enforcing this for all non-sqlite connections by default is a bad thing. Do you see this differently?  
Especially since you integrated the work-around really well into existing configuration mechanisms.

aleyna72072: But I want to make clear that I absolutely lack experience in the administration of DBs.

aleyna72072: Thinking about it, maybe it just does not matter.

doc/installation/system/inifile.rst
```diff
@@ -86,6 +86,28 @@ slower but more robust and can be necessary in large redundant setups.
...
+snapshot is pinned at the transaction's first read, which can cause errors
+like *"Record has changed since last read ...; try restarting transaction"*
+when the same token is used in several concurrent authentication requests.
+PostgreSQL already defaults to ``READ COMMITTED`` and SQLite is not affected,
+so other databases are left untouched.
```
aleyna72072:
```suggestion
PostgreSQL already defaults to ``READ COMMITTED``.
```

SQLite is not a supported database, so imo it does not need to get mentioned here.

aleyna72072 (changes requested):
Beside the one question above, the unittests are failing.  
Other than that LGTM.

edumfa/app.py
```diff
@@ -158,6 +158,22 @@ def create_app(
...
+    # consistent read snapshot at the transaction's first read. During the
+    # concurrent updates that happen while validating the same token from
+    # several requests at once, this raises error 1020 (ER_CHECKREAD,
+    # "Record has changed since last read ...; try restarting transaction").
+    #
+    # PostgreSQL already defaults to READ COMMITTED (which the code targets and is
+    # tested against). SQLite is left untouched as its dialect does
+    # not accept "READ COMMITTED".
+    db_uri = app.config.get("SQLALCHEMY_DATABASE_URI", "") or ""
+    if db_uri.startswith(("mysql", "mariadb")):
```
aleyna72072: Ah, `/tests/test_lib_audit.py` is using EDUMFA_AUDIT_SQL_URI to connect to a separate sqlite for the audit log.  
So outside of the unittest this could mean:
-  "[If EDUMFA_AUDIT_SQL_OPTIONS is not set, SQLALCHEMY_ENGINE_OPTIONS will be used.](https://edumfa.readthedocs.io/en/latest/installation/system/inifile.html#cfgfile)"
- If your audit DB is on a PSQL, and your main one is on a MariaDB, this change would also set the transaction level on the PSQL, which you want to avoid.

Tricky.

fritterhoff:
Ping :)

edumfa/app.py
```diff
@@ -158,6 +158,27 @@ def create_app(
...
+    # MariaDB/MySQL default to the REPEATABLE READ isolation level, which pins a
+    # consistent read snapshot at the transaction's first read. During the
+    # concurrent updates that happen while validating the same token from
+    # several requests at once, this raises error 1020 (ER_CHECKREAD,
+    # "Record has changed since last read ...; try restarting transaction").
+    #
+    # PostgreSQL already defaults to READ COMMITTED (which the code targets and is
+    # tested against). 
+    db_uri = app.config.get("SQLALCHEMY_DATABASE_URI", "") or ""
+    if db_uri.startswith(("mysql", "mariadb")):
```
aleyna72072: I recall @Luc1412 recommending something like `context.get_context().dialect.name in ..` instead. Is that possible here?

doc/faq/mysqldb.rst
```diff
@@ -23,6 +23,14 @@ eduMFA.
 
 **connect string**: ``mysql+pymysql://<user>:<password>@<host>/<database>``
 
+.. note:: MySQL and MariaDB default to the ``REPEATABLE READ`` transaction
+   isolation level, which can lead to errors like *"Record has changed since
```
aleyna72072, line 27: `when the same token is used in several concurrent authentication requests`  
This sounds dangerous without context imo.  
Maybe instead: `when under high load. As security-sensitive operations are done via locking anyway, eduMFA therefore automatically uses the  ``READ COMMITTED`` isolation level.`?

doc/installation/system/inifile.rst
```diff
@@ -86,6 +86,34 @@ slower but more robust and can be necessary in large redundant setups.
...
    might help.
 
+.. _mysql_isolation_level:
+
+If your database is MySQL or MariaDB, eduMFA automatically uses the
+``READ COMMITTED`` transaction isolation level instead of the MySQL/MariaDB
+default ``REPEATABLE READ``. With ``REPEATABLE READ`` a consistent read
+snapshot is pinned at the transaction's first read, which can cause errors
+like *"Record has changed since last read ...; try restarting transaction"*
+when the same token is used in several concurrent authentication requests.
```
aleyna72072: see above

edumfa/app.py
```diff
@@ -158,6 +158,27 @@ def create_app(
...
+        engine_options = dict(app.config.get("SQLALCHEMY_ENGINE_OPTIONS", {}))
+        if "isolation_level" not in engine_options:
+            # The SQL audit module falls back to SQLALCHEMY_ENGINE_OPTIONS if
+            # EDUMFA_AUDIT_SQL_OPTIONS is not set. The audit database can live
+            # on a different database (EDUMFA_AUDIT_SQL_URI), where the injected
+            # isolation level may be invalid (e.g. SQLite), so preserve the
+            # unmodified options as the audit engine's fallback.
+            app.config.setdefault("EDUMFA_AUDIT_SQL_OPTIONS", dict(engine_options))
```
aleyna72072: Why is `engine_options` cast to dict again?

aleyna72072:
Ok, I've thought way too long about this: this is a good change, but I am not too sure about the implementation in the long-term.  
Have you considered setting `engine_options` in https://github.com/eduMFA/eduMFA/blob/v2.9.3/edumfa/models.py#L77 instead? It [is lower priority than SQLALCHEMY_ENGINE_OPTIONS](https://flask-sqlalchemy.readthedocs.io/en/stable/config/#flask_sqlalchemy.config.SQLALCHEMY_ENGINE_OPTIONS), makes the code simpler and then we can leave the tests and the Audit config alone.

edumfa/app.py
fritterhoff: I don't have a context here and must modify the context creation before it is done. -> No real option. The only way would be to create some sort of temporary context, inspect this single variable and then throw the context away again.

fritterhoff:
> Ok, I've thought way too long about this: this is a good change, but I am not too sure about the implementation in the long-term. Have you considered setting `engine_options` in https://github.com/eduMFA/eduMFA/blob/v2.9.3/edumfa/models.py#L77 instead? It [is lower priority than SQLALCHEMY_ENGINE_OPTIONS](https://flask-sqlalchemy.readthedocs.io/en/stable/config/#flask_sqlalchemy.config.SQLALCHEMY_ENGINE_OPTIONS), makes the code simpler and then we can leave the tests and the Audit config alone.

Don't think that will really simplify the code that much but will check in detail. If you want to keep it as simple as possible then you only should add a warning somewhere and fine.

Every other way will require checking the DB type -> Unsure which data exists in the mentioned section.

edumfa/app.py
fritterhoff: Only as a precaution. Maybe not realy required. Wanted to be a bit defensive.

fritterhoff:
So, right now I have no idea how to intercept the DB creation in `models.py`. From my point of view the DB type must be honored since setting the isolation level for sqlite will cause trouble even if it is not an official supported instance. Since we are running the most tests (and also prop. the most dev environments) against sqlite that would cause even more changes.

If that is a major requirement that the change is moved there i will try to figure out some way but if that solves the complexity problem? I am not sure at all.

fritterhoff:
As expected -> Reading the db version like done in https://github.com/eduMFA/eduMFA/pull/1183/changes/2f763c13cf11eb66c622074de62c3418da448a86 does not work out of the box.

fritterhoff:
Added a pretty hacky quirk. Not sure if that is now better than adding the if-else cascade in the `app.py`

aleyna72072:
Ah, I see where the problem is. I assumed this was possible:
```python3
engine_options = None
if is_mariadb:
    engine_options = {
        "isolation_level": "READ COMMITTED",
    }
db = SQLAlchemy(engine_options=engine_options)
```

But I see that is_mariadb is the problem. Bleh :(

aleyna72072:
After some direct chat, these are my preferred solutions:
- The easier way (which would go along with Florian's preferences of not changing the isolation level) would be to require/recommend that setting in #1150: `SQLALCHEMY_ENGINE_OPTIONS = {"isolation_level": "READ COMMITTED"}`. This requires no changes to code, and goes along with Florian's initial proposal of a opt-in.
  + Injecting in the tests can be done via a solution similar to (or even /with) `TEST_DATABASE_URL`.
- More ambitious would be to develop on the basis of PostgreSQL with the help of e.g. devcontainers and to set it as a general default. But this is potentially a larger effort than makes sense at this point.  

I think the first sounds best to me at this point. But if you and @Luc1412 form a consensus, I will obviously be fine with whatever!

fritterhoff:
So. One last attempt. Plz review @Luc1412 and @aleyna72072

aleyna72072 (dismissed):
LGTM

fritterhoff:
Mhm... just noticed the ci is still failing

tests/test_lib_audit.py
```diff
@@ -323,6 +323,10 @@ def test_09_check_external_audit_db(self):
             "edumfa.lib.auditmodules.sqlaudit.create_engine"
         ) as engine_mock:
             cfg = self.app.config.copy()
+            # the app config (e.g. from EDUMFA_CONFIGFILE) may define
+            # EDUMFA_AUDIT_SQL_OPTIONS, which would shadow the fallback
+            # to SQLALCHEMY_ENGINE_OPTIONS we want to test here
+            cfg.pop("EDUMFA_AUDIT_SQL_OPTIONS", None)
```
aleyna72072, line 329: Why is this necessary although `EDUMFA_AUDIT_SQL_OPTIONS = {}` is set?

fritterhoff: as explained in the comment above. In case we set the param the fallback does not work and thus the foo - bar param does not get applied. the `{}` fullfils that sth is set and the copy from the normal engine options does not work.

fritterhoff: (And setting the param is required so the audit is written for the remaining tests...)

aleyna72072: ah I see 👍

aleyna72072 (approved):
LGTM

PR #1197, opened 2026-07-03
feat(image): add support for helm in the entrypoint

aleyna72072:
Since the Helm chart will be in its own repository, this PR is splitting out the entrypoint changes from #1174.  
It should not change existing behaviour*. Only if a variable `CONTAINER_TYPE` is set, the entrypoint does less:
- if "helm_init" simply create/update the db and create/print the admin
- if "helm_worker", wait for the db to be at the latest schema, then start the server

Once this is merged, I can use the `main` tag to use upstream images.  

[*] With one exception: in the Dockerfile, I disabled the control socket. This [allows interacting with gunicorn](https://gunicorn.org/guides/gunicornc/), but requires a write to the filesystem. This is not a documented feature, and users who want it can override the entrypoint.  
It would be nice to have `readOnlyRootFilesystem`, but if this is an issue, it can wait for eduMFA 3.0.0.


Luc1412 (commented):
Changes look good. Just one question: The helm init skips `create_keys` and `execute_user_scripts`. Is this intended?

aleyna72072:
> create_keys

Yes, my initial goal was that the chart can only work with user-provided Kubernetes secret objects.  This makes it so that eduMFA does not need any kind of persistent storage (provider), which simplifies operation imo.  
At some point I probably want to add a kind of demo mode, which enables a Helm install without any values. Maybe then this will change, but I want to do that separately once the chart has proven itself.  

> execute_user_scripts

Yes, I want to add this later as a feature for the migration job, but not for the initial release.

Luc1412 (approved):

PR #1209, opened 2026-07-14
feat: implement monitoring chart

### LINKED ISSUE #1187: feature request dashboard: visualize /monitoring stats
**Background**
PeriodicTasks enable collection of statistics, for example with the included SimpleStats module. This task then (when edumfa-cron is setup) collects different data points (https://github.com/eduMFA/eduMFA/blob/v2.9.3/edumfa/lib/task/simplestats.py#L46), which get then saved to `monitoringstats`. [The values are integers](https://github.com/eduMFA/eduMFA/blob/v2.9.3/edumfa/models.py#L3308).  
The available keys can then be listed on `/monitoring/` and each key can be then requested on `/monitoring/$KEY` (https://edumfa.readthedocs.io/en/v2.9.3/modules/api/monitoring.html).  

The above stats get collected but the evaluation of those values currently has to happen externally, e.g. with SQL queries. A eduMFA-native way to look at the numbers throughout time would be nice.  
This would be helpful and informative during migrations to eduMFA.

An additional widget in the admin dashboard which queries available stats (`GET /monitoring/`) and offers a selector which key to visualize. A graph then gets populated with the numbers from `GET /monitoring/$KEY`. This graph shows the datapoints returned by the endpoint: the values over time.  
Features:

- Some kind of selection which date range to select (e.g. last month, three months, last year, all) would be nice.  
- Depending on the load this puts on the DB/eduMFA, it might make sense to only show a graph/request a key's data once a key has been selected. This would avoid a spike everytime an admin logs on. Maybe the load is negligible in reality though.
- Nice-to-have: If I remember correctly, Chart.js offered a really simple way to show multiple graphs at the same time. Maybe instead of a dropdown with one selection ability, use one where multiple values can be selected. A dropdown like that is used in many places in the WebUI, for example [here](https://github.com/eduMFA/eduMFA/blob/37659d0a7698bab62a991a88acd6f9ebf25b5237/edumfa/static/components/config/views/config.periodictasks.details.html#L74-L89). This would make comparing multiple stats really comfortable.

**Describe alternatives you've considered**
A text-only version in the WebUI or even edumfa-manage or similar. Would also offer a way to look at the numbers, but obviously not as slick.

**Additional notes**
- best to wait for #1146 being merged to avoid merge conflicts later
- A reference how the monitoring endpoint can be queried from the WebUI code and a widget is added can be found in #1146.
- I once used [Chart.js](https://www.chartjs.org/) for a project, the graphs looked really nice but it was also kind of a heavyweight.

**some requirements**

* [ ] A new widget is added (likely `col-lg-12`/entire width at the bottom).
* [ ] The new widget is only visible if the permission is given (`ng-show="checkRight('statistics_read')`).
* [ ] The data points can grow _a lot_ if they are e.g. collected every 5 minutes and are being ran for a few years. I remember Chart.js being a little strange with a lot of data points, but this could be due to misconfig/being long ago. Would be nice if a potential implementer could test this.
* [ ] Request for a key's data is only made when it is selected to avoid load (based on the baseless assumption that this load is even worth mentioning).
* [ ] Test if a graph could be shown by default or if the load is sizeable enough to only show a graph when a key is selected.
* [ ] The code uses the existing code layout/organization as much as possible.

AngelaSchlieben:
Adds a chart visualization for monitoring stats, as requested in #1187.
Changes: 
- New chart component to display monitoring stats
- Integration into the existing dashboard
- Data transformation/preparation for chart rendering


edumfa/static/components/monitoring/factories/monitoring.js
```diff
@@ -0,0 +1,98 @@
...
+                                generateLabels: (chart) => {
+                                    const originalLabels = Chart.defaults.plugins.legend.labels.generateLabels(chart);
+
+                                    return originalLabels.map(label => {
+                                        const isVisible = chart.isDatasetVisible(label.datasetIndex);
+
+                                        if (!isVisible) {
+                                            label.text = '☐ ' + label.text.replace('☐ ', '').replace('☑ ', '');
+                                        } else {
+                                            label.text = '☑ ' + label.text.replace('☐ ', '').replace('☑ ', '');
```
Luc1412: Can we implement a real UI instead of an icon? If there are specific reasons for using unicode character please lmk.

edumfa/static/components/dashboard/views/dashboard.html
```diff
@@ -176,17 +170,42 @@
...
+    </div>
+    <div class="row">
+        <div class="col-lg-12">
+            <div class="panel panel-default" ng-show="checkRight('statistics_read')">
+                <div class="panel-heading">Token Chart</div>
+                <div class="panel-body">
+                    <div class="chart-controls">
+                        <form ng-submit="get_tokens_timeline()">
+                            <div style="display: flex; flex-wrap: wrap; gap: 16px;">
+                                <label>Select TimeFrame:</label>
```
Luc1412:
```suggestion
                                <label>TimeFrame:</label>
```
We don't need to mention the action, since it should be clear with the UI element being present and interactable

edumfa/static/components/dashboard/views/dashboard.html
```diff
@@ -176,17 +170,42 @@
...
+                                <select ng-model="selected_start_time" ng-options="x for x in time_frame"></select>
+                                <label>Select StatsKeys:</label>
```
Luc1412:
```suggestion
                                <label>Stats keys:</label>
```
Same as Timeframe

edumfa/static/components/dashboard/views/dashboard.html
```diff
@@ -176,17 +170,42 @@
...
+                                <isteven-multi-select 
+                                    input-model="stats_keys_liste"
+                                    output-model="selected_stats_keys_objs" 
+                                    button-label="name" 
+                                    item-label="name"
+                                    tick-property="selected">
+                                </isteven-multi-select>
+                                
+                            <button type="submit">Send</button>
```
Luc1412: UX wise I'd prefer a solution with an auto submit. So if a user changed a setting (and it's different to the prev state) we can update the graph.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
```diff
@@ -19,173 +19,250 @@
...
+        };
+
+        $scope.available_stats_keys = [
+            "total_tokens",
+            "hardware_tokens",
+            "software_tokens",
+            "unassigned_hardware_tokens",
+            "assigned_tokens",
+            "user_with_token"
+        ];
```
Luc1412: Is there a reason for it being hard coded? You implemented `get_stats_keys` so it probably is what we want to use here.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
```diff
@@ -19,173 +19,250 @@
...
+            "All",
+            "1 Day",
+            "1 Week",
+            "1 Month",
+            "3 Month",
+            "6 Month",
+            "1 Year"
+        ]
+
+        $scope.selected_start_time = $scope.time_frame[0];
```
Luc1412: `All` being the default timeframe could be bad since it causes a unknowingly large dataset to load with possible a long loading time. It's preferred to preselect a smaller timeframe like 24 Hours or 1 Month

edumfa/static/contrib/js/chart.umd.min.js
Luc1412, line 1: Where do you source the files? 3rd party dependencies need to be added in the `package.json` and `update_contrib.sh` also needs to be updated. This also applied to `chartjs-adapter-date-fns.bundle.min.js`

edumfa/static/components/dashboard/controllers/dashboardControllers.js
```diff
@@ -19,173 +19,250 @@
...
+                    day = day - (x * 7)
+                }
+                else if ($scope.selected_start_time.endsWith("Month")) {
+                    month = month - x
+                }
+                else if ($scope.selected_start_time.endsWith("Year")) {
+                    year = year - x
+                }
+                $scope.start_time = new Date(year, month, day)
+            }
```
Luc1412: The date range calculation is currently a bit hacky. Maybe it's better to update timeframe to an object `{ label, unit, amount }`, add helper function for getting the start date that includes 2 helper functions to subtract year(s) and month(s). So in the end you could call `var startTime = getStartTime($scope.selectedTimeFrame);`

edumfa/static/components/dashboard/controllers/dashboardControllers.js
```diff
@@ -19,173 +19,250 @@
...
+            console.log($scope.start_time)
+
+            $scope.selected_stats_keys = $scope.selected_stats_keys_objs.map(o => o.name);
+            $scope.selected_stats_keys.forEach(sk =>
+                MonitoringFactory.get_monitored(sk, {start: $scope.start_time }, function (data) {
```
Luc1412: Imagine following sequence:
1. The user requests “1 Year.”
2. Before it finishes, they request “1 Day.”
3. The “1 Day” responses arrive and render.
4. A slow “1 Year” response arrives afterward.

This causes issues because every submission replaces `tokens_timeline`. We need to avoid stale asynchronous results by canceling earlier requests e.g. with generation IDs.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
```diff
@@ -19,173 +19,250 @@
...
+        };
+
+        $scope.colorMap = {
+            hardware_tokens: '#4e79a7',
+            software_tokens: '#f28e2b',
+            total_tokens: '#e15759',
+            user_with_token: '#76b7b2',
+            assigned_tokens: '#59a14f',
+            unassigned_hardware_tokens: '#edc948'
+        };
```
Luc1412: Colors should be dynamic. Dynamic/custom keys need a deterministic palette fallback.

edumfa/static/components/dashboard/views/dashboard.html
```diff
@@ -176,17 +170,42 @@
...
+        </div>
+    </div>
+    <div class="row">
+        <div class="col-lg-12">
+            <div class="panel panel-default" ng-show="checkRight('statistics_read')">
+                <div class="panel-heading">Token Chart</div>
+                <div class="panel-body">
+                    <div class="chart-controls">
+                        <form ng-submit="get_tokens_timeline()">
+                            <div style="display: flex; flex-wrap: wrap; gap: 16px;">
```
Luc1412: The div was never closed

edumfa/static/components/dashboard/views/dashboard.html
```diff
@@ -176,17 +170,42 @@
...
+                                <label>Select TimeFrame:</label>
+                                <select ng-model="selected_start_time" ng-options="x for x in time_frame"></select>
```
Luc1412: You are using a normal select here instead of the usual selects you use for stats key. This causes the UI being inconsistent.

Luc1412 (commented):
This is a great first attempt of the implementation. 

I like the UI, but it needs some tweaks for better UX and be more appealing to non-technical users with fewer technical language. We also need a loading, empty and error state.

The implementation also seems solid already. But there are some ruff edges and hacky implementations we can improve. We also maybe need to consider moving code out of `dashboardControllers` into a monitoring controller. But thats something I need to talk to alyena.
For code style you used snake_case. In JS it's a bit different. There you use camelCase for variables and functions.
The new frontend code is missing translations. With the `translate` keyword you can make strings translateable. 

You also included a lot of formatting. This needs to be reverted, since we only want to include relevant changes in a PR.

### TODO
Besides the comments, here a todo list with all important things
- [ ] Reduce technical language
- [ ] Add loading, empty and error state
- [ ] Move Code outside `dashboardControllers`? (need to talk to @aleyna72072)
- [ ] Update code style by using camelCase
- [ ] Revert formatting changes
- [ ] Localize new frontend features
- [ ] Remove debugging logging if finished

edumfa/static/components/monitoring/factories/monitoring.js
AngelaSchlieben: This is specific to chart.js. but instead of the build-in labels, html-elements could be used (though this might make the multiselect for `stats_keys` unnecessary)

edumfa/static/components/dashboard/controllers/dashboardControllers.js
AngelaSchlieben: Just a leftover from an earlier implementation; it will be changed.

edumfa/static/contrib/js/chart.umd.min.js
AngelaSchlieben: https://cdnjs.com/

aleyna72072:
> Move Code outside dashboardControllers? (need to talk to @aleyna72072)

@Luc1412 I am afraid have little knowledge about web development and Angular. The file is currently only the sixth largest controller file, but I absolutely see your concern since it is denser than the other files (which have a lot of function declarations).  
In #1146 I assumed I could have moved some of the code into a directive.  

TLDR: no idea. If you have an idea, please take the lead!

edumfa/static/components/dashboard/controllers/dashboardControllers.js
AngelaSchlieben: Done, but kept the colormap because random generation often produces very similar colors, and without any assignment, the graphs frequently turn out gray—despite using a custom Chart.js color palette.

edumfa/static/components/dashboard/views/dashboard.html
AngelaSchlieben: Done

edumfa/static/components/dashboard/views/dashboard.html
AngelaSchlieben: Done

edumfa/static/components/dashboard/controllers/dashboardControllers.js
AngelaSchlieben: Is now changed.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
AngelaSchlieben: Done

edumfa/static/contrib/js/chart.umd.min.js
AngelaSchlieben: They are now added.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
AngelaSchlieben: Timeframe is now such an object and getStartTime has been adjusted.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
AngelaSchlieben: A check was added to determine whether the dataset is still relevant, and handling for race conditions was implemented to avoid redundant queries.

edumfa/static/components/dashboard/views/dashboard.html
AngelaSchlieben: Done, Stats key select now uses the same CSS class as other selects in the UI

edumfa/static/components/monitoring/factories/monitoring.js
AngelaSchlieben: Removed formatting and added HTML elements to display a legend and to be able to hide the graph.

edumfa/static/components/dashboard/views/dashboard.html
AngelaSchlieben: Switched from a send button to auto-submit and implemented the corresponding logic.

edumfa/static/contrib/js/chart.umd.min.js
Luc1412: Great, but the lockfile is still missing the dependencies. You'd need to run `bun install` once to update the lockfile

edumfa/static/components/dashboard/controllers/dashboardControllers.js
Luc1412: Great addition. 
Unfortunately `All` is currently a bit misleading with it being defined as 100 years. Instead setting unit and amount to null would be better + handling these values.
When it comes to calculation, month subtraction can still overflow at month ends.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
Luc1412: I see caching and relevance checks were added. But the undefined `cacheKey` makes uncached loading fail + Errors also leave pending entries uncleared.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
Luc1412: This solution works, but causes color changes on reload/reset which could be distracting. Instead generating colors randomly, it can be done deterministically from a key. An (accessible) color pallet could be the basis for colors here.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
AngelaSchlieben: Done, the names have been corrected; it should now work as intended.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
AngelaSchlieben: Unit and quantity are now set to "null" and getStartTime() is adjusted accordingly (the API accepts an empty string as a parameter). I tested the calculation with various timeFrames and encountered no overflow issues; the startTime was always correct.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
AngelaSchlieben: Done; additionally, instead of a new color, legend entries that exceed the number of colors in the color scheme now feature a dashed line to make them easier to distinguish.

edumfa/static/contrib/js/chart.umd.min.js
AngelaSchlieben: Done

edumfa/static/components/dashboard/controllers/dashboardControllers.js
```diff
@@ -19,168 +19,373 @@
...
+                        color: colorScheme[i % colorScheme.length],
+                        selected: false,
+                        checked: true,
+                        loadingText: "",
+                        borderDash: getborderDash(i++)
+                    })
+                })
+                $scope.availableStatsKeys = newList
+                if ($scope.availableStatsKeys.length === 0) {
+                    $scope.statsKeysLoadingText = "No available stats keys"
```
Luc1412: Currently this message is just being displayed instead of a chart. I think we need better empty states.

edumfa/static/components/monitoring/factories/monitoring.js
Luc1412: Great improvement! Maybe we just need to find a better to align it properly

edumfa/static/components/dashboard/views/dashboard.html
Luc1412: works great

edumfa/static/components/dashboard/controllers/dashboardControllers.js
Luc1412: Here the issue: For example, subtracting one month from March 31 constructs February 31, which JavaScript normalizes to March 3. The “1 Month” range therefore excludes February entirely. The target day should be clamped to the last valid day of the target month.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
Luc1412: We're almost there. They is only one edge case issue: Failed request leave entries in `pendingRequests` forever, so future requests for that key never run and loading remains stuck. Also, `resetAll()` clears request tracking without cancelling in-flight requests, allowing duplicate requests and unsafe late callbacks.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
Luc1412: Great we getting closer to the final solution, but colors are currently assigned by each key’s position in an unordered API response. If the response order changes, the same stats key can receive a different color after reloading. Assign colors deterministically from the key name or sort the keys first.

edumfa/static/components/monitoring/factories/monitoring.js
AngelaSchlieben: Alignment and overall appearance of the legend are now improved.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
AngelaSchlieben: Using a deferred pattern cancels old requests, and `cancelAll()` can be used to cancel all requests.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
AngelaSchlieben: api response now gets sorted

Luc1412: Sorting makes colors stable while the set of keys remains unchanged. Adding a key earlier alphabetically shifts colors for later keys. Hashing the key into the palette would provide true per-key stability.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
Luc1412: The controller passes error callbacks, but the factory methods do not accept or invoke them.
Consequently, key-list failures still display loading state forever, and dataset failures remain stuck at loading state.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
Luc1412: Great, this is now resolved.

edumfa/static/components/monitoring/factories/monitoring.js
Luc1412: Now it's perfect, great job.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
AngelaSchlieben: Implemented of a solution that maximizes color stability and avoids duplicate color and style assignments while continuing to use the accessible color palette;

edumfa/static/components/dashboard/controllers/dashboardControllers.js
AngelaSchlieben: Errors now get passed; keys should no longer get stuck in the loading state.

edumfa/static/components/monitoring/controllers/monitoringControllers.js
```diff
@@ -0,0 +1,340 @@
...
+        month = month - amount;
+        return calculateDate(year, month, day);
+      }
+      if (unit === "year") {
+        year = year - amount;
+        return calculateDate(year, month, day);
+      }
+    }
+
+    function minimalizeValues(values) {
```
Luc1412, line 114: The reduction protects Chart.js rendering, but it happens after the API has returned and the browser has parsed the complete series. Selecting “All” can therefore still produce an unbounded database query, response, and memory allocation.
Could we add a server-side point limit or aggregation/downsampling parameter? The frontend should request approximately the number of points it can display instead of downloading the complete history.

edumfa/static/components/dashboard/views/dashboard.html
```diff
@@ -256,4 +256,130 @@
...
+        <div class="panel-body" ng-show="availableStatsKeys.length > 0">
+          <div class="chart-controls">
+            <label translate>TimeFrame:</label>
+            <select
+              class="form-control"
+              ng-model="selectedTimeFrame"
+              ng-options="x.label | translate for x in timeFrame "
+              ng-change="onTimeFrameChange()"
+            ></select>
+            <label translate>StatsKeys:</label>
```
Luc1412, line 288: Can we use user-facing terminology here?
- TimeFrame → Time range
- StatsKeys → Metrics
“Stats key” reflects the API terminology, while users primarily need to choose which metric to display.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
Luc1412: The original race condition is mostly resolved now: requests are cancelled, failures release their state, and late dataset results are checked before being added.
Two edge cases remain:
1. When a request is replaced by another request using the same key, the cancelled request’s error callback can clear the replacement request’s loading state.
2. getStatsKeys() is not cancellable or protected by a generation/request ID, so a response started before piReload can still overwrite the newer key state.
Please guard status updates by request identity and apply the same stale-response protection to the stats-key request.

Luc1412: The original race condition is mostly resolved now: requests are cancelled, failures release their state, and late dataset results are checked before being added.
Two edge cases remain:
1. When a request is replaced by another request using the same key, the cancelled request’s error callback can clear the replacement request’s loading state.
2. getStatsKeys() is not cancellable or protected by a generation/request ID, so a response started before piReload can still overwrite the newer key state.
Please guard status updates by request identity and apply the same stale-response protection to the stats-key request.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
Luc1412: This is significantly improved: styles are now derived from the key and there are 64 distinguishable combinations.
One stability edge case remains. Collision resolution uses the currently loaded key set, so adding or removing a colliding key can still change an existing key’s style. If strict per-key stability is required, derive the complete color/dash/point combination directly from the hash without probing for an unused slot.
This is lower priority than the request and large-dataset issues.

edumfa/static/components/dashboard/controllers/dashboardControllers.js
Luc1412: Resolved. The chart now has explicit loading, empty, and error states for both the metric list and individual datasets.

PR #1238, opened 2026-07-15
chore: implement prettier for frontend code formatting

Luc1412:
This PR adds prettier as an dependency and formats the whole frontend code with the tool.


aleyna72072:
~It's a little difficult to tell, does this PR contain functional changes?~

aleyna72072:
Super important PR, hope you can figure out the unit test errors!

Luc1412:
@aleyna72072 All issues are resolved

aleyna72072:
In general this looks good (and honestly every time I look at the UI files I crave for this PR), but it would be way easier to review (and to comprehend in the future) if the syntactic errors and any non-linting changes were a separate PR where possible. Then we could add the linting PR to .git-blame-ignore-revs without hiding possible behaviour changes.  
Since this is just the frontend, I would absolutely understand if this is too much effort at this point. In that case I would just skim and test it out before approving.

aleyna72072 (changes requested):
see above question about splitting this PR

aleyna72072:
Since #1241 is now merged: ping?

aleyna72072 (changes requested):
please rebase/fix the merge conflicts

aleyna72072:
Nice, I applied prettier to main and got the same result as this branch!  
Let's not forget to add the resulting commit to `.git-blame-ignore-revs`.

aleyna72072 (approved):
Thanks!

PR #1241, opened 2026-07-30
fix(webui): correct invalid HTML markup

Luc1412:
## Summary

- fix malformed and mismatched HTML elements across the WebUI templates
- correct duplicate and broken form-control label associations
- fix option labels for machine and token application settings
- preserve valid list, table, and template-fragment structure
- update generated translation catalogs for the corrected DayPassword text

## Verification

- parsed all 157 source HTML files with Prettier
- validated element nesting, duplicate IDs, and permitted content
- ran `git diff --check`


edumfa/static/components/machine/views/machine.details.html
aleyna72072: There are three occurrences of "optionnamme":
```

> git grep optionnamme
edumfa/static/components/machine/views/machine.details.html:                        <label for="option_{{ optionnamme }}">{{ optionname | capitalize }}</label>
edumfa/static/components/machine/views/machine.details.html:            <label for="option_{{ optionnamme }}">{{ optionname | capitalize }}</label>
edumfa/static/components/token/views/token.details.html:                        <label for="option_{{ optionnamme }}">{{ optionname | capitalize }}</label>

```

edumfa/static/components/machine/views/machine.details.html
```diff
@@ -72,20 +72,24 @@ <h3 translate>Tokens and Applications for machine {{ machineid }}</h3>
                     <div class="form-group"
                          ng-repeat="(optionname, option)
                          in Applications[token.application].options">
-                        <label for="option_{{ optionnamme }}">{{ optionname | capitalize }}</label>
+                        <strong id="token_{{ token.id }}_option_{{ optionname }}_label">
```
aleyna72072, line 75: Is it intended that this changed type?

aleyna72072: https://github.com/user-attachments/assets/e1d6eb3f-ab53-4499-89ce-892eb9761a77  
I don't see much of a difference, and I would assume label is correct here?

aleyna72072 (changes requested):
LGTM except the label -> strong change

edumfa/static/components/machine/views/machine.details.html
Luc1412: Yeah, since I use `aria-labelledby`. All 3 inputs existist simultanisly in DOM, while only one gets shown via `ng-show`. But a `label` only can reference one input.

aleyna72072 (approved):
LGTM

PR #1245, opened 2026-08-06
fix: make radiustokens check message-authenticator

arpitjain099:
`RADIUSServer.request` adds a Message-Authenticator when the server is configured with `enforce_ma`, and refuses a response that either omits it or fails verification (`radiusserver.py:123-147`). That is the BlastRADIUS mitigation for CVE-2024-3596.

`RadiusTokenClass._check_radius` builds its own Access-Request and does neither, so a RADIUS server configured with `enforce_ma` still gets an unprotected exchange whenever authentication runs through the token rather than through `RADIUSServer.request`.

This mirrors the existing implementation rather than inventing a second approach:

- resolve `enforce_ma` alongside the other per-path settings, from `radius_server_object.config.enforce_ma` for a configured identifier and from `radius.enforce_ma` for the system-settings path, defaulting to `False` so nothing changes for anyone who has not enabled it
- add the attribute to the request before `PwCrypt`, as `radiusserver.py` does
- verify it on the response *before* the response code is interpreted, so an Access-Challenge cannot bypass the check

**Verification**

`test_12_enforce_message_authenticator` covers all three cases the reference implementation handles: no Message-Authenticator on the response, present and valid, present and broken. The existing `radiusmock` already supports this through `setdata(ma=..., broken_ma=...)`, so no test infrastructure was needed.

Fails on current `main`:

```
FAILED tests/test_lib_tokens_radius.py::RadiusTokenTestCase::test_12_enforce_message_authenticator
1 failed
```

Passes with this change, along with the existing suites:

```
tests/test_lib_tokens_radius.py tests/test_lib_radiusserver.py
24 passed
```

Run on Python 3.12.13.

Opened at @aleyna72072's suggestion on the related advisory. Happy to adjust the approach if you would rather the setting were resolved differently for the individual-token-settings path, which currently defaults to `False` since it has no server object to read from.

### DIFF
--- a/edumfa/lib/tokens/radiustoken.py
+++ b/edumfa/lib/tokens/radiustoken.py
@@ -461,6 +461,7 @@ class RadiusTokenClass(RemoteTokenClass):
         system_radius_settings = self.get_tokeninfo("radius.system_settings")
         radius_timeout = 5
         radius_retries = 3
+        radius_enforce_ma = False
         if radius_identifier:
             # New configuration
             radius_server_object = get_radius(radius_identifier)
@@ -471,10 +472,14 @@ class RadiusTokenClass(RemoteTokenClass):
             radius_dictionary = radius_server_object.config.dictionary
             radius_timeout = int(radius_server_object.config.timeout or 10)
             radius_retries = int(radius_server_object.config.retries or 1)
+            radius_enforce_ma = radius_server_object.config.enforce_ma
         elif system_radius_settings:
             # system configuration
             radius_server = get_from_config("radius.server")
             radius_secret = get_from_config("radius.secret")
+            radius_enforce_ma = get_from_config(
+                "radius.enforce_ma", False, return_bool=True
+            )
         else:
             # individual token settings
             radius_server = self.get_tokeninfo("radius.server")
@@ -528,6 +533,9 @@ class RadiusTokenClass(RemoteTokenClass):
                 NAS_Identifier=nas_identifier.encode("ascii"),
             )
 
+            if radius_enforce_ma:
+                req.add_message_authenticator()
+
             req["User-Password"] = req.PwCrypt(otpval)
 
             if radius_state:
@@ -541,6 +549,22 @@ class RadiusTokenClass(RemoteTokenClass):
                     f"The remote RADIUS server {r_server} timeout out for user {radius_user}."
                 )
                 return AccessReject
+
+            if radius_enforce_ma:
+                # verify_message_authenticator() raises a generic exception
+                # if the M-A attribute is missing, so check for it first
+                if "Message-Authenticator" in response:
+                    if not response.verify_message_authenticator(
+                        original_authenticator=req.authenticator
+                    ):
+                        log.info(
+                            f"Radiusserver {r_server} sent broken "
+                            "Message-Authenticator"
+                        )
+                        return AccessReject
+                else:
+                    log.info(f"Radiusserver {r_server} sent no Message-Authenticator")
+                    return AccessReject
 
             # handle the RADIUS challenge
             if response.code == pyrad.packet.AccessChallenge:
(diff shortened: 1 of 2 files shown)


aleyna72072:
Hi, thanks for the PR - in general, please don't create a public but a private MR (GitHub has even support of this as part of their advisory system), as this breaches the embargo without eduMFA agreeing about this.  
Please keep that in mind for next time (especially if it is actually putting users at risk!).

edumfa/lib/tokens/radiustoken.py
```diff
@@ -471,10 +472,14 @@ def _check_radius(self, otpval, options=None, radius_state=None):
             radius_dictionary = radius_server_object.config.dictionary
             radius_timeout = int(radius_server_object.config.timeout or 10)
             radius_retries = int(radius_server_object.config.retries or 1)
+            radius_enforce_ma = radius_server_object.config.enforce_ma
         elif system_radius_settings:
             # system configuration
             radius_server = get_from_config("radius.server")
             radius_secret = get_from_config("radius.secret")
+            radius_enforce_ma = get_from_config(
```
aleyna72072, line 480: You can remove this line, `system_radius_settings` is a legacy path which should mean this setting can't exist.

aleyna72072 (changes requested):
I can confirm the fix works as intended.

edumfa/lib/tokens/radiustoken.py
```diff
@@ -542,6 +550,22 @@ def _check_radius(self, otpval, options=None, radius_state=None):
...
+                        original_authenticator=req.authenticator
+                    ):
+                        log.info(
+                            f"Radiusserver {r_server} sent broken "
+                            "Message-Authenticator"
+                        )
+                        return AccessReject
+                else:
+                    log.info(f"Radiusserver {r_server} sent no Message-Authenticator")
+                    return AccessReject
```
Luc1412: Returning early misses the option updates for `radius_result`, `radius_state` and `radius_message`. Could this be an issue in `authenticate()` if `radius_result` for example isn't set?

arpitjain099: Yes, it is an issue, and worse than just the missing values. `is_challenge_response` clears `radius_result` at the start of the `check_token_list` loop, so when I bail out before the updates, `authenticate` finds `None` and queries the RADIUS server a second time with the same OTP value.

I counted the packets the mock serves to be sure. Old code, `AccessAccept` without a Message-Authenticator, one options dict through `is_challenge_response` / `is_challenge_request` / `authenticate`:

```
radius_packets=2  radius_result=None  radius_state=None  radius_message=None
```

A plain `AccessReject` from a server without `enforce_ma` on the same chain:

```
radius_packets=1  radius_result=3  radius_state='<REJECTED>'  radius_message='RADIUS authentication failed'
```

So the refusal was not behaving like a rejection at all. Fixed by folding the check into the response-code chain instead of returning: an unverifiable response now sets `result = AccessReject` with the same state and message a rejected authentication gets, and the method runs to its end. Both cases now read `radius_packets=1 radius_result=3`.

The test asserts those three option values now, and also covers a broken Message-Authenticator on an `AccessChallenge`, so a bad packet cannot reach the challenge handling. Reverting the change fails it with `None != 3`.

One thing I left alone: the `Timeout` handler above has the same early return. That one looks intentional to me since a second attempt would just time out again, but say the word if you want it lined up too.

24 tests pass in `test_lib_tokens_radius.py` and `test_lib_radiusserver.py`, ruff check and format clean.

edumfa/lib/tokens/radiustoken.py
arpitjain099: Removed. You are right, `enforce_ma` only exists as a column on the `radiusserver` table, written through the RADIUS server API and the config UI. There is no `radius.enforce_ma` config key anywhere in the tree, so `get_from_config` could only ever hand back the default. The `radius_enforce_ma = False` initialisation above the branch already covers the legacy paths.

Luc1412 (approved):

PR #1256, opened 2026-08-25
feat(image): add engine opts to env based cfg

aleyna72072:
In #1150 we'll recommend/require setting certain engine opts. This patch enables setting those for the container image.  
Adding the isolation level to the example in docker.rst will happen in the above PR.


fritterhoff:
Out of curiosity: would it be possible to add one or two tests for the edumfa_config.py? 🤔

aleyna72072:
> Out of curiosity: would it be possible to add one or two tests for the edumfa_config.py? 🤔

Good idea! I've added test at #1258 (drafted until this is merged since it contains tests for this change, too). First time using an LLM for a contribution, that was exciting (even if a little less useful than expected).

aleyna72072:
@Luc1412 ping?

deploy/docker/edumfa_config.py
```diff
@@ -50,21 +51,41 @@ def get_var(key: str, default: str | None = None) -> str:
...
 
+def str_to_dict(data: str) -> dict:
+    """
+    Parses a string into a dict.
+
+    :param data: The string to convert to a dict.
+    :return: The string parsed as dict.
+    :raises ValueError: If the content of the string is not a dict.
+    """
+    data = literal_eval(data)
```
Luc1412: Environment variables and Docker Compose commonly use JSON. Consider replacing `ast.literal_eval()` with `json.loads()` and documenting JSON as the supported format. This also gives consistent behavior for values such as `true`, `false`, and `null`.

deploy/docker/edumfa_config.py
```diff
@@ -50,21 +51,41 @@ def get_var(key: str, default: str | None = None) -> str:
         return get_content_from_file(path)
     return _getenv(key, default)
 
+def str_to_dict(data: str) -> dict:
```
Luc1412:
```suggestion
def str_to_dict(data: str) -> dict[str, object]:
```

deploy/docker/edumfa_config.py
```diff
@@ -50,21 +51,41 @@ def get_var(key: str, default: str | None = None) -> str:
...
+    Parses a string into a dict.
+
+    :param data: The string to convert to a dict.
+    :return: The string parsed as dict.
+    :raises ValueError: If the content of the string is not a dict.
+    """
+    data = literal_eval(data)
+    data_type = type(data)
+    if data_type != type({}):
+        raise ValueError(f"Expected string with dictionary, got {data_type}.")
```
Luc1412:
```suggestion
    if not isinstance(data, dict):
        raise ValueError(f"Expected string with dictionary, got {data_type}.")
```

deploy/docker/edumfa_config.py
```diff
@@ -50,21 +51,41 @@ def get_var(key: str, default: str | None = None) -> str:
...
+)
 EDUMFA_AUDIT_KEY_PUBLIC = get_var("EDUMFA_AUDIT_KEY_PUBLIC", "/etc/edumfa/public.pem")
 EDUMFA_LOGCONFIG = get_var("EDUMFA_LOGCONFIG", "/opt/edumfa/logging.yml")
 EDUMFA_UI_DEACTIVATED = get_var("EDUMFA_UI_DEACTIVATED", "False") == "True"
 EDUMFA_AUDIT_SQL_TRUNCATE = True
 EDUMFA_NODE = gethostname()
+if sqlalchemy_options := get_var("SQLALCHEMY_ENGINE_OPTIONS", ""):
+    SQLALCHEMY_ENGINE_OPTIONS = str_to_dict(sqlalchemy_options)
+if sqlalchemy_options_audit := get_var("EDUMFA_AUDIT_SQL_OPTIONS", ""):
+    EDUMFA_AUDIT_SQL_OPTIONS = str_to_dict(sqlalchemy_options_audit)
```
Luc1412: If parsing fails during container startup, the current exception does not identify which setting was invalid. Consider passing the variable name into the parser or wrapping each error with context such as `Invalid SQLALCHEMY_ENGINE_OPTIONS`.

doc/installation/docker.rst
```diff
@@ -174,6 +174,8 @@ optional are required for eduMFA to work.
 - EDUMFA_LOGCONFIG: a path to an alternative logging config (optional, default: image provided)
 - EDUMFA_UI_DEACTIVATED: whether to disable the WebUI (optional, default: ``False``)
 - SUPERUSER_REALM: which realms should be superuser realms (optional, default: ``super,administrators``)
+- SQLALCHEMY_ENGINE_OPTIONS: SQLAlchemy engine options, see :ref:`cfgfile` (optional)
+- EDUMFA_AUDIT_SQL_OPTIONS: SQLAlchemy engine options for your audit database, see :ref:`cfgfile` (optional)
```
Luc1412: State whether these values use JSON or Python-literal syntax and provide a short example. Otherwise users may reasonably supply JSON booleans like `true`, which `literal_eval()` rejects.

deploy/docker/edumfa_config.py
Luc1412, line 1: I think this file deserves some testing. Not sure if you want to include it here or in an extra PR.

Luc1412 (changes requested):
Change looks good, but we can optimize a few things.

deploy/docker/edumfa_config.py
aleyna72072: This would create inconsistency between the docker implementation and configuring eduMFA normally. A user would read something in the documentation, try it out with docker (which I would encourage) and get weird failure.

deploy/docker/edumfa_config.py
aleyna72072: https://github.com/eduMFA/eduMFA/pull/1258

deploy/docker/edumfa_config.py
Luc1412: Alright that makes sense.

doc/installation/docker.rst
aleyna72072: See above.  
FWIW, this returns a clear error: `NameError: name 'true' is not defined. Did you mean: 'True'?`

deploy/docker/edumfa_config.py
aleyna72072: As the lines already point to the "inifile" section, I've specified there that this refers to a Python dictionary.

deploy/docker/edumfa_config.py
aleyna72072: fixed

deploy/docker/edumfa_config.py
```diff
@@ -50,21 +51,41 @@ def get_var(key: str, default: str | None = None) -> str:
         return get_content_from_file(path)
     return _getenv(key, default)
 
+def str_to_dict(data: str, option: str) -> dict[str, Any]:
```
Luc1412, line 56: Not sure why you changed `object` to `Any`, but `Any` isn't imported and will probably cause an error.

Luc1412 (changes requested):

deploy/docker/edumfa_config.py
aleyna72072: Didn't throw an error on Python 3.14, let me check with 3.10.

aleyna72072: Ah, lower versions _do_ throw an error. Imported Any.

doc/installation/docker.rst
Luc1412: The intention here was that the user got a straight requirement what data type is expected.

aleyna72072: Which they can now see when looking it up on the cfgfile-link? That seems pretty appropriate, but I'm open for suggestions.  
Note that e.g. `SUPERUSER_REALM` also does not define its data type on that page.

aleyna72072: Oh, SUPERUSER_REALMS actually differs to plain edumfa.cfg, this sucks.  
Added an explanation for both.

aleyna72072: Opened an issue for the difference: https://github.com/eduMFA/eduMFA/issues/1264

Luc1412 (approved):

PR #1258, opened 2026-08-31
chore(tests): add docker config tests

aleyna72072:
Tests the loading of the docker config. Also fixes the CreateLocalCATestCase tearDownClass, which in some cases creates a '.rnd' file which would block removal of the directory.

Assisted by: Mistral Medium 3.5

### DIFF
--- a/tests/test_docker_config.py
+++ b/tests/test_docker_config.py
@@ -1,0 +1,214 @@
+import os
+import sys
+import tempfile
+import unittest
+from pathlib import Path
+from types import ModuleType
+from unittest import mock
+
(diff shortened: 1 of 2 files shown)


aleyna72072:
need to add #1262  

EDIT: This is not happening in the config but the entrypoint script. Added a note.

aleyna72072:
need to add option for str_to_dict

EDIT: done

aleyna72072:
FYI: I have another PR in the pipeline which will test the image. That will test the entrypoint implicitly, too.

PR #1262, opened 2026-09-08
feat(image): add EDUMFA_ADMIN_ENABLE

aleyna72072:
Currently, a local eduMFA admin always gets created in the image. Some users want to disable that to reduce attack surface.  
It also makes the Helm chart configuration a little easier to grasp.


doc/installation/docker.rst
Luc1412, line 184: This section suggests that all vars allow for a `_FILE` suffix. This is not true anymore. We need to adjust this section. But this should be part of another PR.

Luc1412 (approved):

doc/installation/docker.rst
aleyna72072: Agreed, https://github.com/eduMFA/eduMFA/pull/1263

PR #1267, opened 2026-09-18
fix: WebAuthn self-attestation enrollment fails with "Invalid signature received."

belfhi:
## Summary
Enrolling a WebAuthn token that uses packed self-attestation (attStmt contains only alg + sig, no x5c certificate — typical of passkey-style / soft-token authenticators) fails at /token/init with:

edumfa.lib.tokens.webauthn.RegistrationRejectedException: Invalid signature received.

This fix unblocks that enrollment path.

## Root cause
eduMFA verifies WebAuthn enrollment in two phases:

1. The webauthntoken_allowed prepolicy (api/lib/prepolicy.py) calls WebAuthnRegistrationResponse.verify_attestation_statement() without a client_data_hash — a parse-only pass to extract the AAGUID and attestation certificate for the REQ / AUTHENTICATOR_SELECTION_LIST enrollment policies.
2. WebAuthnTokenClass.update() calls .verify(), which invokes the same method with the real client_data_hash — the actual signature check.

verify_attestation_statement() already guards its signature verification with if client_data_hash: in the FIDO_U2F and packed-with-x5c branches, so the parse-only pass correctly skips verification there. The packed self-attestation branch was missing that guard, so it unconditionally verified a signature over authenticatorData || <empty clientDataHash>. A valid attestation signature always covers authenticatorData || clientDataHash, so this can never succeed during the parse-only phase, and every self-attestation enrollment was rejected.

The regression dates to privacyIDEA commit b7d98d6e6 ("Don't require client_data_hash"), which added the guard to the U2F branch but not the self-attestation branch. It only manifests for self-attestation, which is why U2F / basic-attestation tokens were unaffected.

## Fix
Add the same if client_data_hash: guard to the self-attestation branch, mirroring the two sibling branches. This only skips verification in the parse-only phase; the real check still runs in phase 2 via .verify(), so nothing is weakened.

## Changes
- edumfa/lib/tokens/webauthn.py — guard the self-attestation signature verification with if client_data_hash: (with explanatory comment).
- tests/test_lib_tokens_webauthn.py — add test_09e_registration_self_attestation_parse_only_no_hash, which reproduces the prepolicy's parse-only call (no hash) and asserts it returns ATTESTATION_TYPE.SELF_ATTESTATION instead of raising.

## Testing
- pytest tests/test_lib_tokens_webauthn.py → 32 passed (31 existing + 1 new).
- ruff check and ruff format --check → clean.
- Verified against a real failing enrollment payload: the parse-only call now succeeds and returns Self; full verification with the hash still accepts the genuine signature; a corrupted signature with the hash is still rejected.


fritterhoff:
Hi,
thanks for the PR. 

I assume you have configured no `webauthn_authenticator_attestation_level` policy? Using `webauthn_authenticator_attestation_level` with value `none` is working fine for us since approx 2y. 

Just wondering a bit if the PR would allow sending an empty attestation even if we request one?

belfhi:
Hi @fritterhoff I have a policy with 
```
 'webauthn_authenticator_attestation_form': 'indirect',
'webauthn_authenticator_attestation_level': 'none',
```
The token is a Browser Passkey (tested on MacOS with Opera or Edge).
Most other passkeys work. Enrollment as a PrivacyIDEA Passkey token works, too, but they don't use the webauthn.py code for those tokens anymore...

fritterhoff (approved):

fritterhoff:
Ah - got your point. Prop. I will add the test vectors given here as a further verification for the future (https://www.w3.org/TR/webauthn-3/#sctn-test-vectors-packed-self-es256)

